PR #8, opened 2020-01-22
Relicense as MIT

### LINKED ISSUE #7: Licensing question
I just want to open up this point for discussion: The whole recipe stack is licensed as GPLv3+, so far no problems. But what about the patches provided alongside the recipes? I guess most of them are just taken from upstream and published here for convenience reasons, but cannot be simply re-licensed to GPLv3+, or? I guess clarifying the origin and license on a per-patch basis is a must for adoption of this basement layer because a re-licensing of the patches - and thus of the built software (at least in parts) - to GPLv3+ would de-facto prohibit commercial use. We'd need a *maintainable* way of tracking both origin and license on a per-patch basis, though.

Any input and opinions appreciated.

jkloetzke:
The sole use of the GPL-3.0-or-later license prohibited the use of the
Basement project in other projects and left a gray area when patches are
applied to other software that is licensed under an incompatible
license.

There are two important exceptions that are granted beyond the usual
GPL-2.0 (or later) license that now applies to this project:

 * Included patches are covered by the licence of the affected software.
   This exception is taken from Buildroot where it did stand the test of
   time.
 * Using the Basement recipes as a separate layer from another Bob
   project is considered normal use and does not fall under the heading
   of "derived work".

Closes #7.


jkloetzke:
@Ferruck Do you agree with this relicensing? It would be good to get your explicit acknowledgement...

Ferruck:
Don't you think that the mention of `config.yaml` in the exception is a bit too technical and explicit? Seems like it could be a future problem. I'd rather use a more general term there.

Apart from that, regarding the contents of this pull request, I fully agree.

jkloetzke:
> Don't you think that the mention of `config.yaml` in the exception is a bit too technical and explicit? Seems like it could be a future problem. I'd rather use a more general term there.

You are probably right. I will try to come up with something better.

> Apart from that, regarding the contents of this pull request, I fully agree.

:+1:

Ferruck:
After thinking a while about it, how about

```
2. As a special exception, you may create a larger work that references
   the Basement project as a separate recipes layer through the recipe
   layering techniques provided by your build tool. Using part or all of
   the Basement recipes as a separate layer is merely considered normal
   use of the Basement project, and does *not* fall under the heading of
   "derived work".
```

jkloetzke:
I'm not sure if this can be formulated waterproof at all. What's the judicial difference in terms of the copyright in

a) adding some recipe in the basement repository, or
b) creating a project with said recipe and using the basement repository as a layer?

In either case the build result will be the same. From my perspective the amount of "derived work" is also the same. Just the technical realization differs. IMHO this is a contradiction to the original intention. Both cases use the same recipes to produce the same result. Now we either

a) lied to ourself that using the basement project by means of a layer is something different or
b) adding some recipes without modifying the existing ones is not creating a derived work.

I think its hard to argue (b) in the light of the GPL. Reading the LGPL blew my mind by how much is written in section 5 and 6 to specify the "L" in LGPL. Also reading https://lkml.org/lkml/2006/12/17/79 opened my eyes. So I think (a) is actually true.

What does it all mean? I think there are again two possiblities:

1. Scrap the exception and let the user judge what is "derived work" and what not. This may prohibit the usage by some non-open source users but so be it.
2. Allow any usage and move to some permissive license.

While I'm typically inclined to the first option I'm more and more convinced that it doesn't gain anything practically. There is no big open source community for Bob and basement. If we want to foster the adoption of Bob and the basement project then using the MIT or BSD license will be better than adding some half baked exception that nobody really understands.

What do you think?

Ferruck:
First of all, I think we're both out of our profession here, so the following are merely not-that-much-educated guesses.

> What's the judicial difference in terms of the copyright in
>
> a) adding some recipe in the basement repository, or
> b) creating a project with said recipe and using the basement repository as a layer?

Well, in the former case I change other peoples work whereas in the latter I use it.

> In either case the build result will be the same. From my perspective the amount of "derived work" is also the same. Just the technical realization differs. IMHO this is a contradiction to the original intention. Both cases use the same recipes to produce the same result.

You're right that the end result - the *build* result - may be the same, but the way you've used to achieve it differs.

> Now we either
>
> a) lied to ourself that using the basement project by means of a layer is something different or
> b) adding some recipes without modifying the existing ones is not creating a derived work.

I'd say it's not that hard to argue for point *b*, since I'd rather call it an aggregate work (referring to Linus' explanation given in the link you've provided later on). I could easily swap out the basement layer with a layer I've come up on my own or someone else's layer, as long as it provides the same recipes. Compare that with Linus' glibc example. As long as I don't use any glibc specifics, every linkage with the glibc is easily exchangeable with every other POSIX-conformant libc.

And I think that these considerations are well in line with the intention of the LGPL (referring to version [2.1](https://www.gnu.org/licenses/old-licenses/lgpl-2.1) here). First of all, IMHO, using the base layer perfectly fits the first sentence of paragraph 5:

> A program that contains no derivative of any portion of the Library, but is designed to work with the Library by being compiled or linked with it, is called a "work that uses the Library". Such a work, in isolation, is not a derivative work of the Library, and therefore falls outside the scope of this License.

(At least as long as you consider your recipes a program) Additionally, paragraph 6, section b, supports my understanding of aggregate in this context. Of course, these paragraphs aren't in the GPL (I guess neither in version 2.0 nor in version 3.0), but here I'd like to come back to Linus who argues that the creation of the LGPL was just about _comfort_, not about changing any of the intentions of the GPL.

But there's another big point here. You - as the basement layer user - are probably not redistributing neither your nor the basement layer's recipes. IMHO, this is what is widely considered _private use_ and that's compatible with both GPL and LGPL independent on whether we're talking about _aggregate_ or _derived_ work. I think we have to think about the recipes more like we think about a tool like, let's say, the compiler. Would you consider it a GPL violation if you're distributing binaries compiled with a GCC that has in-house modifications?

> What does it all mean? I think there are again two possiblities:
>
> 1. Scrap the exception and let the user judge what is "derived work" and what not. This may prohibit the usage by some non-open source users but so be it.
> 2. Allow any usage and move to some permissive license.

IMHO, point one is a no-option. To keep the wording, licensing is about rules and laws - obviously - but it's also about intentions and comfort. Especially when dealing with multiple international laws that you can't satisfy all at the same time, clearly stating your initial intentions is really important. Point two on the other hand seems like the more secure way for both parties. I fully agree with your following paragraph - I also prefer copyleft licenses - but if we want to gain some traction to Bob, we need to ease adoption. Which non-copyleft license would you prefer and why?

In the end, maybe _software_ licenses simply cannot provide what we need for recipes. Looking at other projects like Bazel and Yocto, the problem is either ignored or not solved for them, either. Have you ever thought about things like CC for the recipes (though I admit that at least CC doesn't solve our _derived_ or _not derived_ problem)?

jkloetzke:
I also looked at CC and I didn't find it a good match either. In the end it all boils down to what is considered "derivative work". And there I think we have to admit that there will be no definitive answer because it is something that a court has to answer in each particular case.

Looking at the permissive licenses I'm in favour of Apache-2.0. It's permissive, keeps away patents as far as possible and requires preservation of the copyright notice and disclaimer. The other possibility would be MIT. In any case there should be an additional clause that patches are under the licence of the patched sources to assure compatibility. This is also done by the Yocto project which is MIT-based.

Ferruck:
> Looking at the permissive licenses I'm in favour of Apache-2.0. It's permissive, keeps away patents as far as possible and requires preservation of the copyright notice and disclaimer.

And to list all changes you've made to the distributed software, IIRC, which can get quite awkward. I've never reasoned about the patent-related parts of the license (are they even relevant here?), but for me the Apache license has always been the sluggish and hard-to-understand sister of the MIT, thus I'd personally prefer the latter. The final decision is up to you, of course.

jkloetzke:
Ok. So it's MIT. No frills...

Ferruck:
Any legal obligations? Do I need to sign that commit of or something? Or is this discussion saved in Github's cloud enough for future record?

jkloetzke:
Hmm, I'll add your signed-off-by and when you explicitly consent here then I think it will be enough. Not sure how you could possibly sign this commit.

Ferruck:
Thanks! Hereby I grant you the irrevocable right to relicense the recipes to the MIT license using the exception mentioned earlier in this thread. :wink:

Issue #26, opened 2020-07-28
classes/cmake: not working getopts 


mahaase:
https://github.com/BobBuildTool/basement/blob/6878d4e8276588e353a7b962a32d9e1d22e8044a/classes/cmake.yaml#L58

I guess this implementation will not work, if u want to use `-i, -m, -o or -O`.
Currently $1 have to be source-path and everything else will be added in the end of the command. So if i use `-i`, i get the error:
`The "cmake -i" wizard mode is no longer supported.`

Maybe we could use something like that. if we need a option for source-path, we have to add this.
usage for this implementation: `cmakeBuild -i install -m all -- -DCMAKE_SOMETHING:BOOL=OFF`

```
diff --git a/classes/cmake.yaml b/classes/cmake.yaml
index 1bbce92..f7372c0 100644
--- a/classes/cmake.yaml
+++ b/classes/cmake.yaml
@@ -3,6 +3,8 @@ inherit: [cpackage, make, install]
 buildTools: [cmake]
 buildVars: [CC, CXX]
 buildScript: |
+    _CMAKE_SOURCE_PATH=$1
+
     # Make sure CMake finds other stuff by its own logic too
     CMAKE_FIND_ROOT_PATH=
     for i in "${@:2}" ; do
@@ -49,16 +51,16 @@ buildScript: |
     {
         local MAKE_TARGET=
         local MAKE_OPTIONS=( )
-        local INSTALL_TAGET=install
+        local INSTALL_TARGET=install
         local INSTALL_OPTIONS=( )

         # parse arguments
-        OPTIND=1
+        local OPTIND=1
         local opt
-        while getopts "i:m:o:O:s" opt ; do
+        while getopts "i:m:o:O:" opt ; do
             case "$opt" in
                 i)
-                    INSTALL_TAGET="$OPTARG"
+                    INSTALL_TARGET="$OPTARG"
                     ;;
                 m)
                     MAKE_TARGET="$OPTARG"
@@ -69,23 +71,26 @@ buildScript: |
                 O)
                     INSTALL_OPTIONS+=( "$OPTARG" )
                     ;;
+                --)
+                    break
+                    ;;
                 \?)
-                  echo "Invalid option: -$OPTARG" >&2
-                  exit 1
-                  ;;
-            esac
+                    echo "Invalid option: -$OPTARG" >&2
+                    exit 1
+                    ;;
+                esac
         done
        shift $(( OPTIND -1 ))

         mkdir -p build install
         pushd build
[...]

Issue #21, opened 2020-07-27
classes/cmake: s option


mahaase:
https://github.com/BobBuildTool/basement/blob/cfc1360a853461b868b8157704225527fe2e6b30/classes/cmake.yaml#L58

why there is a "s" parameter?

jkloetzke:
Good catch. Looks like a copy'n'paste error.

mahaase:
fixed & merged.

PR #27, opened 2020-07-30
classes/cmake: fixed argument handling of cmakeBuild function

mahaase:
fixes for: 
classes/cmake: not working getopts #26
classes/cmake: s option? #21

### DIFF
--- a/classes/cmake.yaml
+++ b/classes/cmake.yaml
@@ -3,6 +3,8 @@ inherit: [cpackage, make, install]
 buildTools: [cmake]
 buildVars: [CC, CXX]
 buildScript: |
+    _CMAKE_SOURCE_PATH=$1
+
     # Make sure CMake finds other stuff by its own logic too
     CMAKE_FIND_ROOT_PATH=
     for i in "${@:2}" ; do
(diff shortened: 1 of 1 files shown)


jkloetzke:
Have you tested this? I've made a quick check and (as I expected) the `--` is already handled by `getopts` automatically. It will also stop parsing options as soon as the first non-option is encountered. So I think the whole patch is not necessary.

The other change in this PR is the removal of the first argument to the sources. This breaks existing recipes that use this class. I would rather make the first argument optional like the `packageScript` functions do it.

mahaase:
for `cmakeBuild $SRCDIR -i install -m all -- -D...` the current implementation will fail like this: 

```
The "cmake -i" wizard mode is no longer supported.
Use the -D option to set cache values on the command line.
Use cmake-gui or ccmake for an interactive dialog.
```

jkloetzke:
Yes, but you just have to switch the order of arguments. All options to `cmakeBuild` must come first, then the path to the sources and then further cmake options: `cmakeBuild -i install -m all -- $SRCDIR -D...`. You can even leave out the `--` because `getopts` will stop parsing options on the first non-option argument...

mahaase:
thx for explanations. it was just a miss-usage, not an issue.

Issue #17, opened 2020-07-27
glibc missing dependency to gawk


mahaase:
missing dependency of util::gawk for glibc build

```
configure: error:
*** These critical programs are missing or too old: gawk
*** Check the INSTALL file for required versions.
++ bob_handle_error 1
```

for compat and non-compat glibc.

I fixed it like that:

```
diff --git a/recipes/libs/compat/glibc.yaml b/recipes/libs/compat/glibc.yaml
index 55c7fe8..49a2ba6 100644
--- a/recipes/libs/compat/glibc.yaml
+++ b/recipes/libs/compat/glibc.yaml
@@ -3,6 +3,9 @@ inherit: [make, patch]
 depends:
     - kernel::linux-libc-headers
     -
+        name: utils::gawk
+        use: [tools]
+    -
         name: devel::compat::gcc-cross-bare
         use: [tools]

@@ -20,7 +23,7 @@ checkoutScript: |
     patchApplySeries $<<glibc/*.diff>>

 buildVars: [AUTOCONF_TARGET]
-buildTools: [gcc-cross-bare, target-toolchain, bison]
+buildTools: [gcc-cross-bare, target-toolchain, bison, gawk]
 buildScript: |
     EXTRA=
     [ -e $1/usr/include/selinux/selinux.h ] || EXTRA+=" --without-selinux"
diff --git a/recipes/libs/glibc.yaml b/recipes/libs/glibc.yaml
index 674cb09..b1a0139 100644
--- a/recipes/libs/glibc.yaml
+++ b/recipes/libs/glibc.yaml
@@ -3,6 +3,9 @@ inherit: [make, patch]
 depends:
     - kernel::linux-libc-headers
     -
+        name: utils::gawk
+        use: [tools]
+    -
         name: devel::gcc-cross-bare
         use: [tools]

@@ -20,7 +23,7 @@ checkoutScript: |
     patchApplySeries $<<glibc/*.diff>>

 buildVars: [AUTOCONF_TARGET]
-buildTools: [gcc-cross-bare, target-toolchain, bison]
+buildTools: [gcc-cross-bare, target-toolchain, bison, gawk]
 buildScript: |
     EXTRA=
     [ -e $1/usr/include/selinux/selinux.h ] || EXTRA+=" --without-selinux"
diff --git a/recipes/utils/gawk.yaml b/recipes/utils/gawk.yaml
index 966cb51..6c79189 100644
--- a/recipes/utils/gawk.yaml
[...]

PR #28, opened 2020-07-30
Feature/added gawk to root

mahaase:
fix for: glibc missing dependency to gawk #17

### DIFF
--- a/classes/basement/rootrecipe.yaml
+++ b/classes/basement/rootrecipe.yaml
@@ -54,3 +54,6 @@ depends:
     - name: utils::e2fsprogs
       use: [tools]
       forward: True
+    - name: utils::gawk
+      use: [tools]
+      forward: True
--- a/recipes/devel/bootstrap-sandbox.yaml
+++ b/recipes/devel/bootstrap-sandbox.yaml
@@ -59,7 +59,6 @@ depends:
     - utils::bzip2
     - utils::file
     - utils::findutils
-    - utils::gawk
     - utils::grep
     - utils::gzip
     - utils::rsync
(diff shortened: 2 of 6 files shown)


jkloetzke:
But this doesn't change anything yet because the tool is not used anywhere. And to enforce consistency the gawk tool should be removed from `recipes/devel/bootstrap-sandbox.yaml` and `recipes/devel/sandbox.yaml` to see if all recipes that require gawk declared so via `buildTools`.

mahaase:
current version fails for sandbox-build

```
configure: error:
*** These critical programs are missing or too old: gawk
*** Check the INSTALL file for required versions.
++ bob_handle_error 1
++ set +x
Step failed with return status 1; Command: $1/configure --prefix=/usr --host=${AUTOCONF_TARGET} --build=$($1/scripts/config.guess) --enable-kernel=4.3.3 --with-headers=$2/usr/include --enable-obsolete-rpc --libdir=/usr/lib $EXTRA
Call stack (most recent call first)
    #0: Recipe libs::glibc#1, layer basement, line 7, in main
>> x86_64::jetson/devel::sandbox/devel::bootstrap-sandbox
   MAP-FPRNT http://fdt-c-pcs-0004.fdtech.intern/artifact/38/07/546a7646a19ebc2a04499c6cca84878a821a-1.fprnt .. not found
Build error: Build script ./dev/build/libs/glibc-dev/1/build.sh returned with 1
Failed package: x86_64::jetson/devel::sandbox/devel::bootstrap-sandbox/devel::cross-toolchain/devel::gcc-cross/libs::glibc-dev
You may resume at this point with '--resume' after fixing the error.
```

any idea what is wrong here?

mahaase:
to build gawk we also need gawk. we have to make gawk to a system requirement.

Issue #24, opened 2020-07-28
cmake: CMAKE_GENERATOR etc. to support windows


mahaase:
https://github.com/BobBuildTool/basement/blob/6878d4e8276588e353a7b962a32d9e1d22e8044a/classes/cmake.yaml#L83

for linux no need for that, but if we want to support windows too, we have to add `CMAKE_GENERATOR, CMAKE_GENERATOR_PLATFORM and CMAKE_GENERATOR_TOOLSET`.

may we should set `CMAKE_GENERATOR: "Unix Makefiles"` in default.yaml.

`CMAKE_GENERATOR, CMAKE_GENERATOR_PLATFORM and CMAKE_GENERATOR_TOOLSET` could be set by special windows toolchain yamls.

jkloetzke:
Sure. Do you plan to send a pull request for this? And related to that shouldn't we then also add some windows toolchain recipes that define these variables?

mahaase:
already merged. we can close this!

THX for that feature!

PR #32, opened 2020-08-07
classes/cmake: MSYS2 Windows support

mahaase:
* added cmake windows support for MSYS2: #24 

**generally we could remove the dependency to the make-class, by using the cmake --build and cmake --install functionality also for linux.** (but we would loose the DESTDIR feature, so we have to check --prefix  and CMAKE_PREFIX_PATH stuff for linux)

feedback welcome!


classes/cmake.yaml
```diff
@@ -19,17 +23,20 @@ buildScript: |
...
-                CMAKE_SYSTEM_NAME=Linux
-                ;;
-            *)
-                CMAKE_SYSTEM_NAME=Generic
-                ;;
-        esac
-        cat >toolchain.cmake <<EOF
+    if [[ ${CMAKE_TOOLCHAIN_FILE:+true} ]] ; then
+        CMAKE_TOOLCHAIN_FILE="${BOB_TOOL_PATHS[target-toolchain]}/${CMAKE_TOOLCHAIN_FILE}"
+    else
```
jkloetzke, line 28: I would rather not re-introduce this historic mistake. The toolchain itself should not bother about the build systems used downstream. If CMake needs a toolchain file then this should be handled by the CMake class. That's the same as for the meson class that also needs something similar.

classes/cmake.yaml
```diff
@@ -1,11 +1,15 @@
...
+buildTools: [cmake, target-toolchain]
+buildVars: [CC, CXX, BOB_HOST_PLATFORM,
+    CMAKE_GENERATOR, CMAKE_GENERATOR_PLATFORM, CMAKE_GENERATOR_TOOLSET,
+    CMAKE_TOOLCHAIN_FILE, CMAKE_BUILD_TYPE, BUILD_SHARED_LIBS]
 buildScript: |
     # Make sure CMake finds other stuff by its own logic too
     CMAKE_FIND_ROOT_PATH=
     for i in "${@:2}" ; do
+        # CMake needs win paths if we are in MSYS2
+        [[ "$BOB_HOST_PLATFORM" == "msys" ]] && i="$(cygpath -m $i)"
```
jkloetzke, line 12: Hmm, isn't that more dependent on the target platform / toolchain? It should still be possible to compile for msys2/mingw23/mingw64 which do not take windows paths I suppose.

classes/cmake.yaml
```diff
@@ -38,10 +45,8 @@ buildScript: |
     set(CMAKE_FIND_ROOT_PATH_MODE_INCLUDE ONLY)
     set(CMAKE_FIND_ROOT_PATH_MODE_PACKAGE ONLY)
     EOF
-        CMAKE_TOOLCHAIN_FILE="$PWD/toolchain.cmake"
-        BUILD_SHARED_LIBS=ON
-    else
-        BUILD_SHARED_LIBS=OFF
```
jkloetzke, line 44: Any reason why you removed this? The rationale is that all natively built libraries should be static because they are presumably used for host tools.

classes/install.yaml
```diff
@@ -11,12 +11,6 @@ inherit: [strip]
 # not pass --static to pkg-config.  Executables are still linked dynamically
 # because glibc cannot be linked statically).
 packageScript: |
-    # Install shared libraries only when cross compiling. On host builds only
-    # static libraries are installed.
-    if [[ "${AUTOCONF_BUILD:-unknown}" != "${AUTOCONF_HOST:-${AUTOCONF_BUILD:-unknown}}" ]]; then
-        INSTALL_SHARED=yes
-    fi
-
```
jkloetzke, line 19: This basically breaks the whole basement Linux build. For host builds no shared libraries must be used because they lead to tools that cannot be executed at all or not reliably.

classes/install.yaml
```diff
@@ -94,10 +88,10 @@ packageScript: |
     {
         installCopy "$@" /usr/ "/usr/include/***" \
             /usr/lib/ \
-            "/usr/lib/*.a" \
-            ${INSTALL_SHARED:+"/usr/lib/*.so*"} \
+            "/usr/lib/***" \
```
jkloetzke, line 91: Now you install shared libraries even if they must not be installed.

classes/install.yaml
```diff
@@ -94,10 +88,10 @@ packageScript: |
...
             "/usr/lib/pkgconfig/***" \
             "/usr/lib/cmake/***" \
+            /usr/bin/ "/usr/bin/***" \
```
jkloetzke, line 94: Why include `/bin` for -dev packages?

classes/install.yaml
```diff
@@ -116,22 +110,19 @@ packageScript: |
...
-        if [[ ${INSTALL_SHARED:-} ]] ; then
-            installCopy "$@" /usr/ /usr/lib/ \
-                "/usr/lib/*.so.*" \
-                "/usr/lib/lib*-*.so" \
-                "!*"
-            installStripAll .
-        fi
+        installCopy "$@" /usr/ \
+            /usr/lib/ "/usr/lib/*.so*" \
+            /usr/bin/ "/usr/bin/*.dll" "/usr/bin/*.pdb" \
```
jkloetzke, line 115: `installPackageLib` should only install shared libraries. By removing `*.so.*` versioned libraries are broken.

classes/install.yaml
```diff
@@ -116,22 +110,19 @@ packageScript: |
...
     # Everything except shared or static libraries or header files.
     installPackageBin()
     {
-        installCopy "$@" '!/usr/include/' \
-            "!/usr/lib/*.a" "!/usr/lib/*.so*"  \
-            "!/usr/lib/pkgconfig" \
-            "!/usr/lib/cmake"
+        installCopy "$@" /usr/ \
+            /usr/bin/ "/usr/bin/***" \
+            "!*"
```
jkloetzke, line 125: What's the reason to change this? Now the function does not install anything except binaries. This will miss e.g. plugins, configuration files, ...

recipes/devel/cmake.yaml
```diff
@@ -68,5 +78,14 @@ multiPackage:
...
             installPackageTgt "$1/install/"
 
+    win64:
+        buildVars: [PKG_VERSION]
+        buildScript: |
+            mkdir -p install/usr
+            rsync -aH --exclude 'doc/' --exclude 'man/' "$1/cmake-${PKG_VERSION}-win64-x64/" install/usr/
+
+        packageScript: |
+            cp -a $1/install/* .
```
jkloetzke, line 88: I think you should put this all into a separate `recipes/devel/cmake-win64.yaml` recipe. The windows CMake recipe has nothing in common with the regular one except the name :laughing:

jkloetzke (changes requested):
I've had a quick glance over the changes.

I think we should first define what the `AUTOCONF_` values are for the MSYS and Windows. They are the central information for all recipes and classes to judge on what platform is compiled (`AUTOCONF_BUILD`) and for which platform the executables/libraries are generated (`AUTOCONF_HOST`) (see [autoconf documentation](https://www.gnu.org/software/autoconf/manual/autoconf-2.65/html_node/Specifying-Target-Triplets.html)).

Quickly googling around lead me to `*-*-mingw32` for native gcc builds and clang seems to use something like `*-win32` for windows. It would be really good to enumerate the target triples. I think the CMake should make the decision about the "generator" based on this information. This decouples the toolchain and CMake class. Other build systems will face similar challenges so it makes sense to have a looser coupling between toolchain and cmake class.

Did you come up with `*-pc-msys` in `plugins/multiarch.py`by yourself or is this already a standard somewhere?

recipes/devel/cmake.yaml
mahaase: I just liked, that this was possible in this way. :D 
But of course, creating a new recipe would have some advantages.
e.g. we can leave the BOB_HOST_PLATFORM stuff away.
I'll fix this.

classes/cmake.yaml
mahaase: Ahh, correct, it doesn't depend directly on MSYS. It's the combination of MSYS and Windows. I have to use the CMAKE_GENERATOR variable. If we use Visual Studio, we are sure, that we are on Windows system. I guess the unix paths should work, if we build e.g. with Unix Makefiles generator.

classes/cmake.yaml
mahaase: Okay. Understand, I'll fix this.

classes/cmake.yaml
mahaase: What I missed here, is that the user can set the type manually inside the recipe.
making a library static or shared should not be limited by the system.
We need a mechanism to overwrite this.

classes/cmake.yaml
```diff
@@ -71,27 +76,73 @@ buildScript: |
...
+        OPTIONS=( ${CMAKE_GENERATOR:+-G"${CMAKE_GENERATOR}"} )
+        if [[ ${CMAKE_GENERATOR} == "Visual Studio"* ]] ; then
+            INSTALL_PREFIX="$DESTDIR$INSTALL_PREFIX"
+            # since VS2019 we can select the platform and target by special variables
+            OPTIONS+=( ${CMAKE_GENERATOR_PLATFORM:+-A"${CMAKE_GENERATOR_PLATFORM}"} )
+            OPTIONS+=( ${CMAKE_GENERATOR_TOOLSET:+-T"${CMAKE_GENERATOR_TOOLSET}"} )
+        else
+            # windows does not support a sigle build type, we have to select the
+            # current type with cmake --build . --config
+            OPTIONS+=( -DCMAKE_BUILD_TYPE=${CMAKE_BUILD_TYPE:-RelWithDebInfo} )
```
mahaase, line 103: Here I'll investigate again, I guess, I also could set the BUILD_TYPE to bob for Windows, if I added this configuration to `CMAKE_CONFIGURATION_TYPES`.

classes/cmake.yaml
```diff
@@ -71,27 +76,73 @@ buildScript: |
...
-            make $INSTALL_TAGET \
-                DESTDIR=${PWD}/../install \
+
+        if [[ ${CMAKE_GENERATOR} == "Visual Studio"* ]] ; then
+            cmake --build . --parallel "${MAKE_JOBS-$(nproc)}" \
+                --config ${CMAKE_BUILD_TYPE:-RelWithDebInfo} \
+                ${MAKE_TARGET:+--target "${MAKE_TARGET}"} \
                 ${INSTALL_OPTIONS:+"${INSTALL_OPTIONS[@]}"}
+        else
+            makeParallel $MAKE_TARGET ${MAKE_OPTIONS:+"${MAKE_OPTIONS[@]}"}
```
mahaase, line 126: What do u think, could we use the cmake --build also for linux? In the end, cmake will call make also.

classes/cmake.yaml
```diff
@@ -71,27 +76,73 @@ buildScript: |
...
+            if [[ ${CMAKE_GENERATOR} == "Visual Studio"* ]] ; then
+                if [[ "$INSTALL_TARGET" == "install" ]] ; then
+                    cmake --install . \
+                        --config ${CMAKE_BUILD_TYPE:-RelWithDebInfo} \
+                        ${INSTALL_OPTIONS:+"${INSTALL_OPTIONS[@]}"}
+                else
+                    # will be built again...
+                    cmake --build . --parallel "${MAKE_JOBS-$(nproc)}" \
+                        --config ${CMAKE_BUILD_TYPE:-RelWithDebInfo} \
+                        --target $INSTALL_TARGET \
```
mahaase, line 138: Do we need a customizable install target? Will this ever used?
The `cmake --install` will call the default install target (cmake_install.cmake).

classes/cmake.yaml
```diff
@@ -71,27 +76,73 @@ buildScript: |
...
+                        ${INSTALL_OPTIONS:+"${INSTALL_OPTIONS[@]}"}
+                fi
+            else
+                make $INSTALL_TARGET \
+                    DESTDIR=$DESTDIR \
```
mahaase, line 143: I will check, if that will change anything in Linux, if we use cmakes install_prefix instead of the DESTDIR of make.

classes/install.yaml
mahaase: I'll check how to merge, that the user could overwrite the library type.

classes/install.yaml
mahaase: cmake will add the lib and the dll in the <package>-config.cmake file automatically. so if the dll is not there, this will break the build.

classes/install.yaml
mahaase: no idea, why i changed this lines handling so linux library files, doesn't depends to the windows cmake stuff. xD. I will revert this.

classes/install.yaml
mahaase: I have to re-check this.

daxcore:
yeah, we have to enumerate the triplets, here is how git for windows project is doing it: https://github.com/git-for-windows/msys2-runtime/commit/6210daaae01a127f9e2c71382bdb965692a6b47b

mahaase:
> I've had a quick glance over the changes.

thanks!

> I think we should first define what the `AUTOCONF_` values are for the MSYS and Windows. They are the central information for all recipes and classes to judge on what platform is compiled (`AUTOCONF_BUILD`) and for which platform the executables/libraries are generated (`AUTOCONF_HOST`) (see [autoconf documentation](https://www.gnu.org/software/autoconf/manual/autoconf-2.65/html_node/Specifying-Target-Triplets.html)).

i am not sure about this. this `AUTOCONF_` stuff comes from the autotools build-tool, doesn't it?
autotools doesn't support windows at all. yeah we could define this as a root-variable for bob, but hmm, at least, choosing the `CMAKE_GENERATOR` isn't a better solution of course :D

> Quickly googling around lead me to `*-*-mingw32` for native gcc builds and clang seems to use something like `*-win32` for windows. It would be really good to enumerate the target triples. I think the CMake should make the decision about the "generator" based on this information. This decouples the toolchain and CMake class. Other build systems will face similar challenges so it makes sense to have a looser coupling between toolchain and cmake class.
> 
> Did you come up with `*-pc-msys` in `plugins/multiarch.py`by yourself or is this already a standard somewhere?

maybe it is a good idea to use the git solution. git is very common, and available for all worlds: https://github.com/git-for-windows/msys2-runtime/commit/6210daaae01a127f9e2c71382bdb965692a6b47b

jkloetzke:
> > I think we should first define what the `AUTOCONF_` values are for the MSYS and Windows. They are the central information for all recipes and classes to judge on what platform is compiled (`AUTOCONF_BUILD`) and for which platform the executables/libraries are generated (`AUTOCONF_HOST`) (see [autoconf documentation](https://www.gnu.org/software/autoconf/manual/autoconf-2.65/html_node/Specifying-Target-Triplets.html)).
> 
> i am not sure about this. this `AUTOCONF_` stuff comes from the autotools build-tool, doesn't it?
> autotools doesn't support windows at all. yeah we could define this as a root-variable for bob, but hmm, at least, choosing the `CMAKE_GENERATOR` isn't a better solution of course :D

It's coming from autotools but they can and should be used elsewhere too. They are used as central switches to define the host and target environment. If all classes consistently use these variables then there is hopefully no need to have extra variables for the different build systems.

> > Quickly googling around lead me to `*-*-mingw32` for native gcc builds and clang seems to use something like `*-win32` for windows. It would be really good to enumerate the target triples. I think the CMake should make the decision about the "generator" based on this information. This decouples the toolchain and CMake class. Other build systems will face similar challenges so it makes sense to have a looser coupling between toolchain and cmake class.
> > Did you come up with `*-pc-msys` in `plugins/multiarch.py`by yourself or is this already a standard somewhere?
> 
> maybe it is a good idea to use the git solution. git is very common, and available for all worlds: [git-for-windows/msys2-runtime@6210daa](https://github.com/git-for-windows/msys2-runtime/commit/6210daaae01a127f9e2c71382bdb965692a6b47b)

Yes, this looks promising. We just have to come up with some triples for VS builds. Maybe `x86_64-pc-win32`? This also nicely reflects the fact that you are actually cross-compiling in your MSYS environment: from `x86_64-pc-msys` to `x86_64-pc-win32`.

BTW, IIRC CMake finds the installed visual studio automatically, right? Or do you have a real Visual Studio toolchain recipe? I have the vague feeling that when building with VS the build results are dependent on the VS version, right?

It would be good to have an example project that showcases the MSYS usage. Otherwise it's hard to tell if we're on the right track.

mahaase:
> Yes, this looks promising. We just have to come up with some triples for VS builds. Maybe x86_64-pc-win32? This also nicely reflects the fact that you are actually cross-compiling in your MSYS environment: from x86_64-pc-msys to x86_64-pc-win32.

We have to different between building with bob natively or in MSYS2 environment, havn't we?
x86_64-pc-win32 ... sounds like native windows msvc build.
x86_64-pc-msys ... the same like x86_64-pc-win32, but in MSYS2

> BTW, IIRC CMake finds the installed visual studio automatically, right? Or do you have a real Visual Studio toolchain recipe? I have the vague feeling that when building with VS the build results are dependent on the VS version, right?

here i invested some time. strange stuff IMO. if we use cmake without -G (generator string), than cmake detects automatically the newest MSVC. If there is a special requirement to the compiler version we have to set this.

First: I guess we should JUST support Visual Studio 2019. VS2019 supports VS compiler versions of VS2015 and VS2017 by setting a toolset version. new since VS2019 is the following syntax: `cmake -G"Visual Studio 16 2019" -A x64 -T v142`

**CMAKE_GENERATOR: "Visual Studio 16 2019"** --> FIX
CMAKE_GENERATOR_PLATFORM: "x64"
CMAKE_GENERATOR_TOOLSET: "v142"

x64: 64 bit
x86: 32 bit

v140: VS2015 compiler version
v141: VS2017 compiler version
v142: VS2019 compiler version

A problem. If we have a non-cmake module, like boost or qt4.
boost just requires the VS prompt (cmd), there are the env-vars `INCLUDE, LIB, LIBPATH and PATH` are set.
qt4 uses the b2, this one also requires this environment variables.

Problem, it is not possible to load this environment variables in MSYS2, because the `vcvarsall.bat` files could not executed successfully in MSYS2. there is a deprecated overview of paths inside the env-vars: https://renenyffenegger.ch/notes/Windows/development/Visual-Studio/environment-variables/index 

i am searching for a solution to get `INCLUDE, LIB, LIBPATH and PATH` values by cmake by specifing the generator. but this doesn't match with the idea to select everything by `AUTOCONF_` value.

Another problem: the compiler itself inside the VS version can be updated by M$ VS-Installer.
e.g.: VS2019 has installed on my PC currently the following versions: 14.22.27905, 14.23.28105, 14.25.28610
all versions > 14.16 are VS2019 compilers..

Next problem, just VS installation is just one part. the second part is the windows kit (sdk).
e.g. VS2010 uses the SDK v7.1. VS2019 uses v10. but here also there are different versions like: 10.0.16299, 10.0.17134, .... 
normally CMake uses the newest one... 

next problem: if i load the `vcvarsall.bat` in windows by regular way, and call CMake without a generator, than also the newest versions are selected and not this ones, I loaded by `vcvarsall.bat` 

example for vcvars: `vcvarsall.bat x64 -vcvars_ver=14.25`

conclusion: windows and toolchains and env-vars is strange stuff.

our current solution is the following: 
```
provideVars:
    CMAKE_GENERATOR: "Visual Studio 16 2019"
    CMAKE_GENERATOR_PLATFORM: "x64"
    CMAKE_GENERATOR_TOOLSET: "v142"

    # Visual Studio 2019 Win64 compiler - default installation paths
    # use "?" char for latest version
    VS_INCLUDE: "C:/Program Files (x86)/Microsoft Visual Studio/2019/Professional/VC/Tools/MSVC/?/include;C:/Program Files (x86)/Windows Kits/10/Include/?/shared;C:/Program Files (x86)/Windows Kits/10/Include/?/um;C:/Program Files (x86)/Windows Kits/10/Include/?/ucrt"
    VS_LIB: "C:/Program Files (x86)/Microsoft Visual Studio/2019/Professional/VC/Tools/MSVC/?/lib/x64;C:/Program Files (x86)/Windows Kits/10/Lib/?/um/x64;C:/Program Files (x86)/Windows Kits/10/Lib/?/ucrt/x64"
    VS_LIBPATH: "C:/Program Files (x86)/Microsoft Visual Studio/2019/Professional/VC/Tools/MSVC/?/lib/x64"
    VS_PATH: "/c/Program Files (x86)/Microsoft Visual Studio/2019/Professional/VC/Tools/MSVC/?/bin/HostX64/x64:/c/Program Files (x86)/Windows Kits/10/bin/?/x64"
```

all developers in the company have to use the default installation paths. we use VS2019 professional. for the problem, that the compiler-version can be updated automatically by M$, we use a bob-class, which search for the "?" char, goes into the directory, and selects the highest number and replaces this with the "?". xD that we export the VS_* vars to MSYS2 without the VS_

jkloetzke:
Wow. This Windows stuff is so f***** insane. :sob: I have to think about it.

Generally speaking the `AUTOCONF_` variables won't be used to select a particular visual studio version. They are just to select the general host/target. Normally the CC/LD/... variables from the toolchain do the rest. But this is apparently different on Windows.

mahaase:
i found something very interesting. if i build boost with the b2 tool, the tool will generate a "msvc-setup.bat" file. 
in the file are the msvc paths defined for `INCLUDE, PATH, LIBPATH and LIB`. The batchfile is such simple, so it also can be loaded by MSYS2.
e.g. if i call `./b2.exe address-model=64 toolset=msvc-14.2` it get 
`bin.v2\standalone\msvc\msvc-14.2\address-model-64\architecture-x86\msvc-setup.bat`
the content in the end, is the same like `vcvarsall.bat `would do under native windows.
but until now, i found no way to extract this from the full boost package.

jkloetzke:
This is a total mess. I have the feeling that the settings must be somehow gathered by a plugin which calls `vcvarsall.bat` (or even `vswhere.exe` which seems to have a standard location) to fetch the values. The toolchain recipe calls the plugin to extract the right values and then provides them.

But CMake always seems to need its own settings. If it can't be done otherwise then we have to live with that. Interestingly the Visual Studio installation already comes with CMake.

Meson and other seem to use the variables provided by `vcvarsall.bat`. At least there seems to be some hope.

mahaase:
yeah, but we need a way to source the `vcvarsall.bat`. it will not work out of the box in MSYS2. b2 (boost build) calls a cmd instance, in which the bat will called, followed by the build command - both in the same cmd instance.

jkloetzke:
I think we must spawn a cmd instance that calls vcvarsall and prints all variables. The output is then parsed by the plug-in.

mahaase:
what do u think, should we implement this from the scratch by using `vswhere` and calling the `vcvarsall.bat` or should we use the b2 (boost build) tool, to get this functionality. in case of b2, we just have to convert the `msvc-setup.bat` to bash-syntax.

jkloetzke:
> what do u think, should we implement this from the scratch by using `vswhere` and calling the `vcvarsall.bat` or should we use the b2 (boost build) tool, to get this functionality. in case of b2, we just have to convert the `msvc-setup.bat` to bash-syntax.

I think it should be implemented without any unnecessary external dependencies. I was thinking about a plugin that executes a single batch file that does the following:

* execute `vswhere.exe` from the known location to find the toolset that we're interested in, 
* call `vcvarsall.bat` to fetch environment,
* print environment to stdout.

The plugin can then parse stdout and get all the variables we're interested in.

> we also need a mapping. b2 selects just compiler versions. 10.0 - 14.2.
> in CMake u have to know, where ur compiler is installed.
> ...
> EDIT: like i said:
> > First: I guess we should JUST support Visual Studio 2019

Yes. I think that's reasonable. Regarding the toolset version there still needs to be a special handling for CMake. All other build systems just use the environment variables.

The `CMAKE_GENERATOR_PLATFORM` can be inferred from `AUTOTOOLS_HOST`. I think the same can be done for the `CMAKE_GENERATOR`. Normally nothing is passed except for `$AUTOTOOLS_HOST = *-win32` where VS2019 is used.

mahaase:
plugin py-file:

```
from os.path import join
import subprocess

VSWHERE = "C:/Program Files (x86)/Microsoft Visual Studio/Installer/vswhere.exe"

WHITELIST = ["INCLUDE", "LIB", "LIBPATH", "PATH"]

def normalizeString(string):
    return string.strip().decode("utf-8")

def vsvars(args, **options):
    r = normalizeString(subprocess.check_output([VSWHERE, '-property', 'installationPath', '-version', '[16.0,17.0)']))
    r = subprocess.check_output([join(r, 'VC/Auxiliary/Build/vcvarsall.bat'), 'x64', '&&', 'set'])
    l = r.splitlines()
    result=[]
    for i in l:
        for j in WHITELIST:
            if normalizeString(i).startswith(j + "="):
                result.append(normalizeString(i))
    return ', '.join(result)

manifest = {
    'apiVersion' : "0.15",
    'stringFunctions' : {
        "vsvars" : vsvars
    },
}
```

maybe we should make `-version [16.0,17.0)` to an option and `x64` must be an option.

the interpreter class:

```
environment:
    VS_VARS: "$(vsvars)"

buildVars: [VS_VARS]
buildScript: |
    convert()
    {
        OLDIFS=$IFS
        IFS=';'
        for j in $1 ; do
            list=${list:+$list:}$(cygpath -u $j)
        done
        IFS=$OLDIFS
        echo $list
    }

    VS_VARS=${VS_VARS//, /,}
    OLDIFS=$IFS
    IFS=','
    for i in $TEST ; do
        envvar=${i%%=*}
        value=${i//$envvar=/}
        list=$(convert $value)

        old=$(printenv $envvar) || true
        if [[ "${old:-false}" != "false" ]] ; then
            if [[ $old == *";"* ]] ; then
                old=$(convert $old)
            fi
        fi
        export "$envvar=$list${old:+:$old}"
    done
    IFS=$OLDIFS
```

Is there a way to minimize the logging-output if i'm using `-vv`?
because of the loops the output is huge.

just optimized for MSYS2, will not work for native windows, maybe we should adapt this, too.

the line with `printenv || true` is a bit strange, but i got no better solution, if there is no variable in environment, that could be printed.

feedback welcome :)

jkloetzke:
I've tested a bit and I don't think you have to convert the paths in your bash script. Just gather all variables in the plugin, cache them and query them one by one (*not* tested):

```python
import os, os.path
import subprocess

cache = {}

def vsvars(args, **options):
    varname = args[0]
    version = args[1]
    platform = args[2:]
    vswhere = os.path.join(os.environ["ProgramFiles(x86)"], "Microsoft Visual Studio/Installer/vswhere.exe")

    tag = (version, tuple(platform))
    if tag not in cache:
        r = subprocess.check_output([vswhere, '-property', 'installationPath', '-products', '*',
                '-requires', 'Microsoft.VisualStudio.Component.VC.Tools.x86.x64', '-version', version],
            universal_newlines=True).strip()
        r = subprocess.check_output([os.path.join(r, 'VC/Auxiliary/Build/vcvarsall.bat')] +
                platform + ['&&', 'set'],
            universal_newlines=True)
        env = {}
        for l in r.splitlines():
            k,sep,v = l.strip().partition("=")
            if k == 'PATH':
                # convert to POSIX path to be mergeable
                v = subprocess.check_output(["cygpath", "-u", "-p", v], universal_newlines=True).strip()
            env[k] = v
        cache[tag] = env

    return env[varname]
```

Then in the toolchain you can:

```yaml
environment:
    "INCLUDE" : "$(vsvars,INCLUDE,'[16.0,17.0)',x64)"
    "VS_PATH" : "$(vsvars,PATH,'[16.0,17.0)',x64)"
```

The only thing that needs special care is `PATH`. It is converted to a POSIX path by the plugin. This has to be appended manually to `$PATH` in the `buildScript`. The other variables can be taken as-is AFAICT.

I noticed that it takes quite a long time to fetch the variables. I think they should be cached somehow on disk. I'm just not sure what would invalidate them if the user updates, installs or removes toolsets.

mahaase:
reviewed and tested. small issue:

```
-    return env[varname]
+        return env[varname]
+    return cache[tag][varname]
```

toolchain file:
```
privateEnvironment:
    PLATFORM: "x64"
    VERSION: "[16.0,17.0)"

provideVars:
    CMAKE_GENERATOR: "Visual Studio 16 2019"
    CMAKE_GENERATOR_PLATFORM: "${PLATFORM}"
    CMAKE_GENERATOR_TOOLSET: "v142"

    INCLUDE: "$(vsvars,INCLUDE,${VERSION},${PLATFORM})"
    LIBPATH: "$(vsvars,LIBPATH,${VERSION},${PLATFORM})"
    LIB: "$(vsvars,LIB,${VERSION},${PLATFORM})"
    VS_PATH: "$(vsvars,PATH,${VERSION},${PLATFORM})"
    
    CPPFLAGS: ...
```

class to activate environment:
```
buildVars: [INCLUDE, LIB, LIBPATH, VS_PATH]
buildScript: |
    export INCLUDE=$INCLUDE
    export LIB=$LIB
    export LIBPATH=$LIBPATH
    export PATH=$PATH:$VS_PATH
```
important to bring `PATH` in front of `VS_PATH`.

to support more than vs2019, we just have to replace `VC/Auxiliary/Build/vcvarsall.bat` string.
we have to search for `vcvarsall.bat` in `VC`. will work for >= vs2015 compilers (?installed by VS2019?).

mapping:
```
"[14.0,15.0)" --> vc140 --> compiler of VS2015
"[15.0,16.0)" --> vc141 --> compiler of VS2017
"[16.0,17.0)" --> vc142 --> compiler of VS2019
```

```
import glob
...
        r = subprocess.check_output([vswhere, '-legacy', '-property', 'installationPath', '-version', version],
            universal_newlines=True).strip()
        for filename in glob.glob(os.path.join(r, 'VC/**/vcvarsall.bat'), recursive = True):
            p = filename
            break
        r = subprocess.check_output([p] + platform + ['&&', 'set'],
            universal_newlines=True)
```

`-products` and `-requires` isn't important IMO. so we can activate `-legacy`.

just 1 more thing... could it a good idea, to give `vsvars` function the `vc142` string and the function maps this to the range string `"[16.0,17.0)"`?

jkloetzke:
> reviewed and tested. small issue:

:+1: 

> toolchain file:

Just a side note: while the usual `provideVars` will still work it is recommended to follow the new style to attach the toolchain specific variables directly to the toolchain. You can have a look at `recipes/devel/cross-toolchain.yaml` to see how it's used. This is useful in two respects: the variables will only be set if you really use the toolchain and it will enable to use the toolchain remapping. So it basically should look like:

 ```yaml
    target-toolchain:
        path: "just_a_placeholder"
        environment:
            CMAKE_GENERATOR: "Visual Studio 16 2019"
            ...
```

> class to activate environment:

No need to re-export variables unchanged. Also make sure to use quotes and append only if `VS_PATH` is set:

```yaml
buildVars: [INCLUDE, LIB, LIBPATH, VS_PATH]
buildScript: |
    ${VS_PATH:+export PATH="$PATH:$VS_PATH"}
```

> to support more than vs2019, we just have to replace `VC/Auxiliary/Build/vcvarsall.bat` string.

According to the Microsoft docs this path is valid for VS 2017 and 2019. Do we really have to support ancient versions too? I thought 2019 should be enough.

> ```python
>         r = subprocess.check_output([vswhere, '-legacy', '-property', 'installationPath', '-version', version],
>             universal_newlines=True).strip()
> ```
> 
> `-products` and `-requires` isn't important IMO. so we can activate `-legacy`.

Not sure. I only installed the build tools and not the whole VS which should be enough to compile C/C++ code. For that you need the `-products` and `-requires` switches to find it. I would like to keep that. It would make sense, e.g. on a Jenkins slave.

> just 1 more thing... could it a good idea, to give `vsvars` function the `vc142` string and the function maps this to the range string `"[16.0,17.0)"`?

No, I don't think this is a good idea. Maybe someone wants to specify a more specific version in the future? Better keep the flexibility.

mahaase:
could we make the legacy-mode optionally?

`if arg[3]: True than -legacy else:  '-products', '*', '-requires', 'Microsoft.VisualStudio.Component.VC.Tools.x86.x64'`
need help for implementation.

everything else look very nice, i would say. 👍 

btw: how u bring color into the code-blocks :)

mahaase:
NOTE to me: 
open tasks: adapt to `AUTOCONF_`

jkloetzke:
I've just added a `tests/linux` directory with some small smoke tests for CMake and meson. I would really like to get this extended and have a `tests/msys` which builds the same packages on Windows. That way we have some smoke tests where we can see that the Windows changes are working as expected and that nothing else breaks.

mahaase:
I added the plugin and the toolchain file in current version. That's all I have so far.
Let me know... :)

jkloetzke:
:+1: I'll hopefully look into it this evening.

jkloetzke:
Ok, I've looked into it this weekend and could simplify it quite a bit. The VS-Generator is not necessary if the environment is setup correctly. The result is here: https://github.com/jkloetzke/basement/tree/feature/cmakeWindowsSupport I've added the test case to `tests/msys`. This is far from complete and the LInux build is now broken.

I also tried to get meson working but this does not seem to work at all. Meson does not work in an msys environment. It feels like we will have to add Python as 3rd language to Bob so that we can build true cross-platform recipes...

mahaase:
```bash
[[ "$AUTOCONF_HOST" == *-win32 ]] && i="$(cygpath -m $i)"
```
I guess, it should by `*-msys`.

```bash
            *-win32)
                CMAKE_SYSTEM_NAME=Windows
                ;;
```
is `*-msys` missing? But .... do we need them? who needs cross-compiling for windows?

The changes you made for calling cmake-configure, build and install, I do not understand :-/

the cpackage-class I do not understand so far. ... and is pkg-config supported for windows?

and what is meson? never heard about that before, should i know it?

jkloetzke:
> > The VS-Generator is not necessary if the environment is setup correctly.
>
> What do u mean with VS-generator?

The CMake "Visual Studio 16 2019" generator. This generator does some magic to find the toolchain itself. No other build system does that. But if the environment is setup correctly then the "Ninja" CMake generator can be used directly and CMake behaves like other build systems too.

> ```shell
> [[ "$AUTOCONF_HOST" == *-win32 ]] && i="$(cygpath -m $i)"
> ```
> 
> I guess, it should by `*-msys`.

No. We build for `x86_64-pc-win32` (AUTOCONF_HOST) on `x86_64-pc-msys` (AUTOCONF_BUILD).

> ```shell
>             *-win32)
>                 CMAKE_SYSTEM_NAME=Windows
>                 ;;
> ```
> 
> is `*-msys` missing? But .... do we need them? who needs cross-compiling for windows?

You do in fact cross compile. Your build system, including Bob, runs on msys. This is a different ABI than what you are compiling for: win32 - native windows binaries without any msys dependency.

> the cpackage-class I do not understand so far. ... and is pkg-config supported for windows?

This is the common class for all c/c++ based packages. And no, pkg-config is not supported on Windows. :grinning: 

> and what is meson? never heard about that before, should i know it?

This is just another build system that gains traction. It is like autotools and CMake. I used it as a litmus test to see if the changes go in the right direction, i.e. that they are general enough to cover not only CMake. Even if it is not a use case for you it should be possible to compile packages other than CMake too on Windows. But because meson and msys are not compatible it was deemed to fail...

mahaase:
> This is far from complete and the LInux build is now broken.

I build the `linux+freertos` demo project on my native linux pc:

Build result is in dev/dist/demo/linux+freertos/1/workspace
Duration: 0:27:50.791080, 27 checkouts (0 overrides active), 57 packages built, 1 downloaded.

jkloetzke:
> > This is far from complete and the LInux build is now broken.
> 
> I build the `linux+freertos` demo project on my native linux pc with my feature branch:

This project does not use CMake. I've pushed an update for my branch that fixes the CMake class and cleans it further up. I still have to fix the meson build of the tests on Linux but it's going in a direction which I'm comfortable with. Please have a look.

> Which build-systems are supported in windows and MSYS. I just know CMake?! Meson supports Windows, but not MSYS2? What is the reason there - i guess it can just be an issue with the path-strings.

Currently it's CMake but meson can be used on Windows too. Just not for msys directly.

> works for VS2019 compiler, but not for VS2015 compiler. there is an issue in the vcvarsall.bat files. the sdk 10.0 will not loaded fully, so rc.exe and mt.exe are missing. I love that windows stuff...
> If u search online for this bug, u will fine a ton of sites, with exactly this issue: e.g. [xmake-io/xmake#225 (comment)](https://github.com/xmake-io/xmake/issues/225#issuecomment-429210442)

Can't we just apply the workaround in the `vsvars` plugin too? Looks like there is "just" something missing in `%PATH%`..?

mahaase:
```cmake
CMAKE_INSTALL_PREFIX="/usr"
```

is there a reason, why my u doesn't used my solution:

```cmake
CMAKE_INSTALL_PREFIX="\/"
```

the backslash prevents the converting to C:/msys64/usr and the usr is not necessary, refer here:
https://cmake.org/cmake/help/latest/module/GNUInstallDirs.html?#special-cases

> For <dir> other than the SYSCONFDIR, LOCALSTATEDIR and RUNSTATEDIR, the value of CMAKE_INSTALL_<dir> is prefixed with usr/ if it is not user-specified as an absolute path. For example, the INCLUDEDIR value include becomes usr/include. This is required by the GNU Coding Standards

if we prefix the `/usr`, `SYSCONFDIR` will be `/usr/etc`.

But, we have to make CMAKE_INSTALL_PREFIX variable. i would prefer: 

```cmake
CMAKE_INSTALL_PREFIX="${CMAKE_INSTALL_PREFIX:-\/}"
```

so a module, that doesn't support `GNUInstallDirs`, has the option to set a path to e.g. `${PWD}/../install/usr`.
But than we have to remove the `--prefix` argument.

___

```bash
[[ "$AUTOCONF_HOST" == *-win32 ]] && i="$(cygpath -m $i)"
```
what is with native windows? it will not support cygpath tool. shouldn't we check `[[ "$AUTOCONF_BUILD == *-msys ]]`.

___

what is about this:

```cmake
-DCMAKE_PREFIX_PATH="/usr/lib/cmake;/usr"
```

for find_package we use this CMake install path structure: `<prefix>/(lib/<arch>|lib|share)/cmake/<name>*/ (U)` this one is just supported for UNIX. For Windows we can use this: `<prefix>/<name>*/ (W)` by adding the `CMAKE_PREFIX_PATH` `/usr/lib/cmake`. For UNIX we add `/usr`.

___


```bash
DESTDIR="${PWD}/../install" cmake \
    ${INSTALL_COMPONENT:+-DCOMPONENT="$INSTALL_COMPONENT"} \
    -P cmake_install.cmake
```

why not using the common CMake way? 
```bash
        if [[ -n "$INSTALL_TARGET" ]] ; then
            if [[ "$INSTALL_TARGET" != "install" ]] ; then
                COMPONENT="--component $INSTALL_TARGET"
            fi
            cmake --install . \
                --config ${CMAKE_BUILD_TYPE:-Bob} \
                --prefix ${CMAKE_INSTALL_PREFIX:-"../install"} \
                ${COMPONENT:+${COMPONENT}} \
                ${INSTALL_OPTIONS:+"${INSTALL_OPTIONS[@]}"}
```

___

Ninja:

the MSYS2 ninja is broken, im my tests. 

the self-built ninja wasn't used. 
so we have to add the dependency somewhere? something like the rootrecipe in linux?!

but, the self-built ninja also not working in MSYS2:

```bash
-- The C compiler identification is MSVC 19.25.28614.0
-- The CXX compiler identification is MSVC 19.25.28614.0
-- Check for working C compiler: /c/Program Files (x86)/Microsoft Visual Studio/2019/Professional/VC/Tools/MSVC/14.25.28610/bin/HostX64/x64/cl.exe
-- Check for working C compiler: /c/Program Files (x86)/Microsoft Visual Studio/2019/Professional/VC/Tools/MSVC/14.25.28610/bin/HostX64/x64/cl.exe - broken
CMake Error at /usr/share/cmake-3.17.3/Modules/CMakeTestCCompiler.cmake:60 (message):
  The C compiler

    "/c/Program Files (x86)/Microsoft Visual Studio/2019/Professional/VC/Tools/MSVC/14.25.28610/bin/HostX64/x64/cl.exe"

  is not able to compile a simple test program.

  It fails with the following output:

    Change Dir: /home/mhaase/ws/bob-example-embedded/dev/build/devel/ninja/x64_64-pc-win32/1/workspace/build/CMakeFiles/CMakeTmp

    Run Build Command(s):/usr/bin/ninja.exe cmTC_4c0a4 && [1/2] Building C object CMakeFiles/cmTC_4c0a4.dir/testCCompiler.c.obj
    Microsoft (R) C/C++-Optimierungscompiler Version 19.25.28614 f▒r x64
    Copyright (C) Microsoft Corporation. Alle Rechte vorbehalten.

    cl : Befehlszeile warning D9024 : Unbekannter Typ der Quelldatei "C:/msys64-1/nologo", Objektdatei wird angenommen.
    cl : Befehlszeile warning D9027 : Quelldatei "C:/msys64-1/nologo" wird ignoriert.
    cl : Befehlszeile warning D9024 : Unbekannter Typ der Quelldatei "C:/msys64-1/DWIN32", Objektdatei wird angenommen.
    cl : Befehlszeile warning D9027 : Quelldatei "C:/msys64-1/DWIN32" wird ignoriert.
    cl : Befehlszeile warning D9024 : Unbekannter Typ der Quelldatei "C:/msys64-1/D_WINDOWS", Objektdatei wird angenommen.
    cl : Befehlszeile warning D9027 : Quelldatei "C:/msys64-1/D_WINDOWS" wird ignoriert.
    cl : Befehlszeile warning D9024 : Unbekannter Typ der Quelldatei "C:/msys64-1/Zi", Objektdatei wird angenommen.
    cl : Befehlszeile warning D9027 : Quelldatei "C:/msys64-1/Zi" wird ignoriert.
    cl : Befehlszeile warning D9024 : Unbekannter Typ der Quelldatei "C:/msys64-1/Ob0", Objektdatei wird angenommen.
    cl : Befehlszeile warning D9027 : Quelldatei "C:/msys64-1/Ob0" wird ignoriert.
    cl : Befehlszeile warning D9024 : Unbekannter Typ der Quelldatei "C:/msys64-1/Od", Objektdatei wird angenommen.
    cl : Befehlszeile warning D9027 : Quelldatei "C:/msys64-1/Od" wird ignoriert.
    cl : Befehlszeile warning D9024 : Unbekannter Typ der Quelldatei "C:/msys64-1/RTC1", Objektdatei wird angenommen.
    cl : Befehlszeile warning D9027 : Quelldatei "C:/msys64-1/RTC1" wird ignoriert.
    cl : Befehlszeile warning D9024 : Unbekannter Typ der Quelldatei "C:/msys64-1/showIncludes", Objektdatei wird angenommen.
    cl : Befehlszeile warning D9027 : Quelldatei "C:/msys64-1/showIncludes" wird ignoriert.
    cl : Befehlszeile warning D9027 : Quelldatei "C:/msys64-1/FoCMakeFiles/cmTC_4c0a4.dir/testCCompiler.c.obj" wird ignoriert.
    cl : Befehlszeile warning D9024 : Unbekannter Typ der Quelldatei "C:/msys64-1/FdCMakeFiles/cmTC_4c0a4.dir/", Objektdatei wird angenommen.
    cl : Befehlszeile warning D9027 : Quelldatei "C:/msys64-1/FdCMakeFiles/cmTC_4c0a4.dir/" wird ignoriert.
    cl : Befehlszeile warning D9024 : Unbekannter Typ der Quelldatei "C:/msys64-1/FS", Objektdatei wird angenommen.
    cl : Befehlszeile warning D9027 : Quelldatei "C:/msys64-1/FS" wird ignoriert.
    [2/2] Linking C executable cmTC_4c0a4.exe
    FAILED: cmTC_4c0a4.exe
    : && /usr/bin/cmake.exe -E vs_link_exe --intdir=CMakeFiles/cmTC_4c0a4.dir --rc="/c/Program Files (x86)/Windows Kits/10/bin/10.0.19041.0/x64/rc.exe" --mt="/c/Program Files (x86)/Windows Kits/10/bin/10.0.19041.0/x64/mt.exe" --manifests  -- "/c/Program Files (x86)/Microsoft Visual Studio/2019/Professional/VC/Tools/MSVC/14.25.28610/bin/HostX64/x64/link.exe" /nologo CMakeFiles/cmTC_4c0a4.dir/testCCompiler.c.obj  /out:cmTC_4c0a4.exe /implib:cmTC_4c0a4.lib /pdb:cmTC_4c0a4.pdb /version:0.0  /machine:x64  /debug /INCREMENTAL /subsystem:console  kernel32.lib user32.lib gdi32.lib winspool.lib shell32.lib ole32.lib oleaut32.lib uuid.lib comdlg32.lib advapi32.lib && :
    RC Pass 1: command "/c/Program Files (x86)/Windows Kits/10/bin/10.0.19041.0/x64/rc.exe /fo CMakeFiles/cmTC_4c0a4.dir/manifest.res CMakeFiles/cmTC_4c0a4.dir/manifest.rc" failed (exit code 1) with the following output:
    Microsoft (R) Windows (R) Resource Compiler Version 10.0.10011.16384
    Copyright (C) Microsoft Corporation.  All rights reserved.

    fatal error RC1107: invalid usage; use RC /? for Help
    ninja: build stopped: subcommand failed.

  CMake will not be able to correctly generate this project.
Call Stack (most recent call first):
  CMakeLists.txt:3 (project)

-- Configuring incomplete, errors occurred!
```

**I don't get Ninja up in MSYS2.**

btw: for building Ninja we use Ninja.

mahaase:
there is another problem. against my hope... the vswhere.exe tool doesn't show the installation path of the _VS compiler_, it shows the path to the _VS installation_.

so if i ask for version "[15.0,16.0)" i don't get the location of the _vs2017 compiler_, which was installed by _VS2019_.

so my toolchain file is wrong, if we just want to support _VS2019 installation_ with included compilers of vs2015, vs2017 and vs2019.

the legacy mode is just to support older _VS installations_.

the resulting problem: I have to call all the time the same vcvarsall.bat file. so how i can choose the compiler version **without** setting the `CMAKE_GENERATOR_TOOLSET` variable?

cc, cxx = cl will always select ... hmm? maybe the newest vs2019 compiler?

and just setting the `CMAKE_GENERATOR_TOOLSET` will also not work in combination of `CMAKE_GENERATOR=Ninja`.
so we are limited to the newest installed compiler version if we are using Ninja?

mahaase:
**conclusion**: 

defaults:
* support VS2019 with latest compiler
* default `CMAKE_GENERATOR` is "Ninja"
* variable `CMAKE_GENERATOR` for overwriting from recipe
* variable/optional `CMAKE_GENERAOTR_TOOLSET` and `CMAKE_GENERAOTR_PLATFORM` for overwriting from recipe
* we just need `vswhere -version [16.0,17.0)`, but keep it variable for overwriting/reusing by different recipes
* default is "x64" - NOTE there is no `CMAKE_GENERAOTR_PLATFORM` "x86", it is called "Win32" 🙄 
* vs2019-toolset.yaml should not provide multiPackages, but we can change to default by using: 
```yaml
- depends: 
  - name: devel::vs2019-toolchain
    environment:
      CMAKE_GENERATOR: <...>
      CMAKE_GENERATOR_*: <...>
    use|forward:<...>
```

mahaase:
maybe u could live with that?

```patch
diff --git a/classes/cmake.yaml b/classes/cmake.yaml
index 2516cac..0a81724 100644
--- a/classes/cmake.yaml
+++ b/classes/cmake.yaml
@@ -1,24 +1,18 @@
 inherit: [cpackage, ninja, install]

 buildTools: [cmake]
-buildVars: [AUTOCONF_HOST, CC, CXX, BUILD_SHARED_LIBS]
+buildVars: [CMAKE_GENERATOR, CMAKE_GENERATOR_PLATFORM, CMAKE_GENERATOR_TOOLSET,
+    CMAKE_INSTALL_PREFIX, CMAKE_BUILD_TYPE,
+    AUTOCONF_HOST, AUTOCONF_BUILD, CC, CXX, BUILD_SHARED_LIBS]
 buildScript: |
     # Make sure CMake finds other stuff by its own logic too
     CMAKE_FIND_ROOT_PATH=
     for i in "${@:2}" ; do
         # CMake needs win paths if we are in MSYS2
-        [[ "$AUTOCONF_HOST" == *-win32 ]] && i="$(cygpath -m $i)"
+        [[ "$AUTOCONF_BUILD" == *-msys ]] && i="$(cygpath -m $i)"
         CMAKE_FIND_ROOT_PATH+="${CMAKE_FIND_ROOT_PATH:+;}$i"
     done

-    # This looks odd but it prevents MSYS from converting /usr to C:\msys64\usr
-    # (or wherever MSYS was installed).
-    if [[ "$AUTOCONF_HOST" == *-win32 ]] ; then
-        CMAKE_INSTALL_PREFIX="C:/usr"
-    else
-        CMAKE_INSTALL_PREFIX="/usr"
-    fi
-
     # CMake does not honor CPPFLAGS! Merge them with C(XX)FLAGS.
     if [ "${CPPFLAGS:+true}" ] ; then
         CFLAGS+=" ${CPPFLAGS}"
@@ -96,22 +90,50 @@ buildScript: |
         mkdir -p build install
         pushd build

+        if [[ ${CMAKE_GENERATOR:-} == "Visual Studio"* ]] ; then
+            # since VS2019 we can select the platform and target by special variables
+            OPTIONS+=( ${CMAKE_GENERATOR_PLATFORM:+-A"${CMAKE_GENERATOR_PLATFORM}"} )
+            OPTIONS+=( ${CMAKE_GENERATOR_TOOLSET:+-T"${CMAKE_GENERATOR_TOOLSET}"} )
+            OPTIONS+=( -DCMAKE_CONFIGURATION_TYPES:STRING="Debug;Release;RelWithDebInfo;MinSizeRel;Bob" )
+        else
+            # windows does not support a sigle build type, we have to select the
+            # current type with cmake --build . --config
+            OPTIONS+=( -DCMAKE_BUILD_TYPE:STRING=${CMAKE_BUILD_TYPE:-Bob} )
+        fi
+
+        # for find_package we use this cmake install path structure:
+        # <prefix>/(lib/<arch>|lib|share)/cmake/<name>*/ (U)
+        # this one is just supported for UNIX. For Windows we can use this:
+        # <prefix>/<name>*/ (W)
+        # by adding the CMAKE_PREFIX_PATH /usr/lib/cmake. For UNIX we add /usr.
         cmake "$1" \
-            -G Ninja \
+            -G"${CMAKE_GENERATOR:-Ninja}" \
             ${CMAKE_TOOLCHAIN_FILE:+-DCMAKE_TOOLCHAIN_FILE="$CMAKE_TOOLCHAIN_FILE"} \
             -DCMAKE_FIND_ROOT_PATH="$CMAKE_FIND_ROOT_PATH" \
-            -DCMAKE_BUILD_TYPE=Bob \
-            -DCMAKE_INSTALL_PREFIX="$CMAKE_INSTALL_PREFIX" \
+            -DCMAKE_PREFIX_PATH="/usr/lib/cmake;/usr;/" \
+            -DCMAKE_INSTALL_PREFIX=${CMAKE_INSTALL_PREFIX:-"\/"} \
             -DBUILD_SHARED_LIBS=$BUILD_SHARED_LIBS \
+            ${OPTIONS:+"${OPTIONS[@]}"} \
             "${@:2}"

-        ninjaParallel ${MAKE_OPTIONS:+"${MAKE_OPTIONS[@]}"} \
-            ${MAKE_TARGETS:+"${MAKE_TARGET[@]}"}
+        if [[ "${CMAKE_GENERATOR:-Ninja}" == "Ninja" ]] ; then
+            ninjaParallel ${MAKE_OPTIONS:+"${MAKE_OPTIONS[@]}"} \
+                ${MAKE_TARGETS:+"${MAKE_TARGET[@]}"}
+        else
+            cmake --build . --parallel "${MAKE_JOBS-$(nproc)}" \
+                --config ${CMAKE_BUILD_TYPE:-Bob} \
+                ${MAKE_TARGETS:+--target "${MAKE_TARGETS[@]}"} \
+                ${MAKE_OPTIONS:+"${MAKE_OPTIONS[@]}"}
+        fi

         if [[ -n "$INSTALL" ]] ; then
-            DESTDIR="${PWD}/../install" cmake \
-                ${INSTALL_COMPONENT:+-DCOMPONENT="$INSTALL_COMPONENT"} \
-                -P cmake_install.cmake
+            if [[ "$INSTALL_COMPONENT" != "install" ]] ; then
+                COMPONENT="--component $INSTALL_COMPONENT"
+            fi
+            cmake --install . \
+                --config ${CMAKE_BUILD_TYPE:-Bob} \
+                --prefix ${CMAKE_INSTALL_PREFIX:-"${PWD}/../install"} \
+                ${COMPONENT:+${COMPONENT}}
         fi
         popd
     }
diff --git a/recipes/devel/vs2019-toolchain.yaml b/recipes/devel/vs2019-toolchain.yaml
index 4fe71ab..b633cb4 100644
--- a/recipes/devel/vs2019-toolchain.yaml
+++ b/recipes/devel/vs2019-toolchain.yaml
@@ -10,12 +10,20 @@ packageVars: [PLATFORM, VERSION]
 packageScript: |
     echo "$PLATFORM $VERSION" > version.txt

+privateEnvironment:
+    VERSION: "[16.0,17.0)"
+    PLATFORM: "x64"
+
 provideTools:
     target-toolchain:
         path: "just_a_placeholder"
         environment:
             AUTOCONF_HOST: "x64_64-pc-win32"

+            CMAKE_GENERATOR: "${CMAKE_GENERATOR:-}"
+            CMAKE_GENERATOR_PLATFORM: "${CMAKE_GENERATOR_PLATFORM:-}"
+            CMAKE_GENERATOR_TOOLSET: "${CMAKE_GENERATOR_TOOLSET:-}"
+
             INCLUDE: "$(vsvars,INCLUDE,${VERSION},${PLATFORM},${LEGACY:-False})"
             LIBPATH: "$(vsvars,LIBPATH,${VERSION},${PLATFORM},${LEGACY:-False})"
             LIB: "$(vsvars,LIB,${VERSION},${PLATFORM},${LEGACY:-False})"
@@ -28,42 +36,3 @@ provideTools:
             CFLAGS: "-O$(if-then-else,$(eq,${BASEMENT_OPTIMIZE},0),d,${BASEMENT_OPTIMIZE})$(if-then-else,${BASEMENT_DEBUG}, -Zi,) -W3"
             CXXFLAGS: "-O$(if-then-else,$(eq,${BASEMENT_OPTIMIZE},0),d,${BASEMENT_OPTIMIZE})$(if-then-else,${BASEMENT_DEBUG}, -Zi,) -W3"
             LDFLAGS: ""
-
-multiPackage:
-    vc140:
-        privateEnvironment:
-            VERSION: "[14.0,15.0)"
-            LEGACY: "True"
-
-        multiPackage:
-            x64:
-                privateEnvironment:
-                    PLATFORM: "x64"
-            x86:
-                privateEnvironment:
-                    PLATFORM: "x86"
-
-    vc141:
-        privateEnvironment:
-            VERSION: "[15.0,16.0)"
-            LEGACY: "True"
-
-        multiPackage:
-            x64:
-                privateEnvironment:
-                    PLATFORM: "x64"
-            x86:
-                privateEnvironment:
-                    PLATFORM: "x86"
-
-    vc142:
-        privateEnvironment:
-            VERSION: "[16.0,17.0)"
-
-        multiPackage:
-            x64:
-                privateEnvironment:
-                    PLATFORM: "x64"
-            x86:
-                privateEnvironment:
-                    PLATFORM: "x86"
diff --git a/tests/linux/layers/self b/tests/linux/layers/self
index 1b20c9f..36dc815 120000
--- a/tests/linux/layers/self
+++ b/tests/linux/layers/self
@@ -1 +1 @@
-../../../
\ No newline at end of file
+../../../
diff --git a/tests/msys/recipes/cmake/greeter.yaml b/tests/msys/recipes/cmake/greeter.yaml
index ce05c1c..bce5a88 100644
--- a/tests/msys/recipes/cmake/greeter.yaml
+++ b/tests/msys/recipes/cmake/greeter.yaml
@@ -5,7 +5,7 @@ depends:
     - name: devel::cmake-win64
       use: [tools]
       forward: True
-    - name: devel::vs2019-toolchain-vc142-x64
+    - name: devel::vs2019-toolchain
       use: [environment, tools]
       forward: True
     - cmake::libgreet-dev
diff --git a/tests/msys/recipes/meson/greeter.yaml b/tests/msys/recipes/meson/greeter.yaml
index 98bd4cd..f82c333 100644
--- a/tests/msys/recipes/meson/greeter.yaml
+++ b/tests/msys/recipes/meson/greeter.yaml
@@ -5,7 +5,7 @@ depends:
     - name: devel::meson
       use: [tools]
       forward: True
-    - name: devel::vs2019-toolchain-vc142-x64
+    - name: devel::vs2019-toolchain
       use: [environment, tools]
       forward: True
     - meson::libgreet
```

waruqi:
> Which build-systems are supported in windows and MSYS. I just know CMake?! Meson supports Windows, but not MSYS2? What is the reason there - i guess it can just be an issue with the path-strings.

xmake can also be obtained and used on windows/msys.

jkloetzke:
> maybe u could live with that?

To be honest: no. :smile: I don't want to treat CMake in any way special. And I still think it is not necessary. I've updated the vs2019-toolchain and the v141 and v142 toolchain versions do work here now. I didn't bother to test v140. I'm pretty sure it works too. I also added ninja to the `basement::rootrecipe` class which now selects the right tools based on the host platform.

I also moved the tools to `recipes/devel/win/...` to keep them separated a bit.

> note: do not use static paths for cmake, ${CMAKE_INSTALL_LIBDIR} instead of lib. so cmake willl automatically prefix the /usr if necessary.

Why? Using a relative path will let CMake add the prefix. It compiles on Linux and Windows(msys) here. Why should we ever use `${CMAKE_INSTALL_LIBDIR}`? CMake will search all paths on all platforms. The (W) and (U) indicators are suggestions AFAICT. I deliberately created the tests from the [cmake documentation](https://cmake.org/cmake/help/latest/manual/cmake-packages.7.html). If some of your modules need to set `CMAKE_PREFIX_PATH` then it should be done in the project, possibly by a dedicated class...

mahaase:
first of all... wow. impressive performance 👍 

just 3 small issues..

___

```patch
-AUTOCONF_HOST: "x64_64-pc-win32"
+AUTOCONF_HOST: "x86_64-pc-win32"
```

___

and with the changes in the "rootrecipe", the "bob_example_embedded" has the issue now, that "host-toolchain" will be needed, but isn't available. for linux i guess the compat or sandbox-toolchain will do this?!

___

again to the "/" vs "(C:)/usr" prefix:
https://cmake.org/cmake/help/latest/module/GNUInstallDirs.html?#special-cases

In my test, i used: `install(FILES doc/mainpage.txt TYPE SYSCONF)` and this results in:

`-- Installing: C:/msys64-1/home/mhaase/ws/bob-example-embedded/dev/build/fdt/example/lib-dev/x86_64-pc-win32/1/workspace/build/../install/usr/etc/mainpage.txt`

`/usr/etc` is not correct for GNU paths IMO, but if i just use the "/", the "usr" will be added automatically by cmake, but it will not be removed automatically. but this just seems to work, if we just use the `CMAKE_INSTALL_*DIR` variables and not e.g. "lib" directly. 🙄 

so the only cross-platform and working solution is the backslash-slash variant as far as i know:

`CMAKE_INSTALL_PREFIX="\/"`

one cool thing, it works for linux and windows.

mahaase:
push. i am back from vacation :)
what's with my last comments?

jkloetzke:
I haven't looked too deeply into it. I have a number of changes pending that will collide with this PR. Once they have landed I will pick up here...

jkloetzke:
I've directly updated the pull request with the latest changes. I hope everything is addressed. Please test...

mahaase:
hey! thanks for the great work. nearly everything looks very good in my tests.

The only thing is the scenarion to use: `install(FILES doc/MainPage.md TYPE SYSCONF)` with the expectation to install into `/etc` directory.

But current result is `/usr/etc`:
```
-- Installing: C:/msys64-1/home/mhaase/ws/bob-example-embedded/dev/build/fdt/example/lib-dev/x86_64-pc-win32/1/workspace/build/../install/usr/etc/MainPage.md
~/ws/bob-example-embedded/dev/build/fdt/example/lib-dev/x86_64-pc-win32/1/workspace
```

It is also possible to use the `//` instead of `//usr`, if `-DCMAKE_PREFIX_PATH="/;/usr"` is added.

why u want the `/usr` for `CMAKE_INSTALL_PREFIX`. In my test, it is not possible to install something to `/etc` or `/var` (in CMake: SYSCONFDIR or LOCALSTATEDIR)

yes, in case of `/` for the `CMAKE_INSTALL_PREFIX` we have to use for install destinations the CMake GNUInstallDirs variables.

otherwise a `DESTINATION lib` would be installed to `/lib` not to `/usr/lib`.
a `DESTINATION ${LIBDIR}` will install into `/usr/lib` automatically.

In the end it is up to you, that is just how I understood CMakes GNUInstallDirs.
Your implementation has the disadvantage, that it isn't possible to install something not into `/usr` directory

jkloetzke:
I think the "/usr" prefix should be kept on Linux builds. OTOH a prefix is somehow useless on Windows. In any case the prefix should be consistent throughout the project. If I understand correctly you only need to set `CMAKE_PREFIX_PATH` if different prefix paths are used simultaneously in the same project. I would not like to support that, though.

I find it surprising that `install(FILES doc/MainPage.md TYPE SYSCONF)` does not do the right thing. Maybe we should set `CMAKE_INSTALL_SYSCONFDIR` to the right absolute path (`/etc`) unconditionally. That should solve the problem. I did not test if this works, though.

mahaase:
okay I'm fine! 👍 

indeed, this works: `-DCMAKE_INSTALL_SYSCONFDIR="//etc" \` for MSYS2.

mahaase:
soooo, when this will go live? :)

jkloetzke:
The meson build of the example needs to be fixed and I have to test here if it breaks anything else. I hope somewhen next week...

jkloetzke (approved):

PR #38, opened 2020-12-12
gstreamer, meson, glib

rhubert:
- a few changes to the meson class
- updated glib to latest version
- added gstreamer recipes

### DIFF
--- a/classes/meson.yaml
+++ b/classes/meson.yaml
@@ -26,12 +26,54 @@ buildScript: |
         unset CXXFLAGS
         unset LDFLAGS
 
+        case "$AUTOCONF_HOST" in
+            *-msys | *-win32)
+                MESON_SYSTEM=windows
+                ;;
+            *-linux-*)
+                MESON_SYSTEM=linux
+                ;;
+            *)
+                echo "Unsupported system: $AUTOCONF_HOST" >&2
+                exit 1
+                ;;
+        esac
+
+        case "$AUTOCONF_HOST" in
+            aarch64-*)
+                MESON_CPU_FAMILY=aarch64
+                : ${MESON_CPU=armv8}
+                ;;
+            arm-*)
+                MESON_CPU_FAMILY=arm
+                : ${MESON_CPU=armv7}
+                ;;
+            i[34567]86-*)
+                MESON_CPU_FAMILY=x86
+                : ${MESON_CPU=i686}
+                ;;
+            x86_64-*)
+                MESON_CPU_FAMILY=x86_64
+                : ${MESON_CPU=x86_64}
+                ;;
+            *)
+                echo "Unsupported cpu family: $AUTOCONF_HOST" >&2
+                exit 1
+                ;;
+        esac
+
         cat >cross_file.txt <<EOF
+    [host_machine]
+    system = '$MESON_SYSTEM'
+    cpu_family = '$MESON_CPU_FAMILY'
+    cpu = '$MESON_CPU'
+    endian = 'little'
     [binaries]
     c     = '$CC'
     cpp   = '$CXX'
     ar    = '$AR'
     strip = '$STRIP'
+    pkgconfig = '$(which pkg-config)'
     [properties]
     c_args = [$(meson_join_list "${MESON_CFLAGS[@]}")]
     cpp_args = [$(meson_join_list "${MESON_CXXFLAGS[@]}")]
--- a/recipes/multimedia/gstreamer.yaml
+++ b/recipes/multimedia/gstreamer.yaml
@@ -1,0 +1,40 @@
+inherit: [meson, patch]
+
+metaEnvironment:
+   PKG_VERSION: "1.18.2"
+
+depends:
+    # Requires glib-mkenums
+    - name: libs::glib-tools
+      use: [tools]
+      tools:
+        target-toolchain: host-toolchain
+
+    - libs::glib-dev
+    - use: []
+      depends:
+        - libs::glib-tgt
+
+checkoutSCM:
+    scm: url
+    url: https://gstreamer.freedesktop.org/src/gstreamer/gstreamer-${PKG_VERSION}.tar.xz
+    digestSHA1: 2a417e7c5eb184b705e0964bb63f83f1498bdf84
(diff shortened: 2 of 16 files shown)


classes/meson.yaml
```diff
@@ -52,9 +56,14 @@ buildScript: |
...
             declare -a MESON_OPTIONS=( "$1" "$PWD")
             if [[ -e $BOB_CWD/cross_file.txt ]] ; then
+
+                if [[ "${MESON_FILE:-undefined}" != "undefined" ]]; then
+                    cat ${BOB_TOOL_PATHS['target-toolchain']}/${MESON_FILE} >> ../$cross_native
+                fi
                 MESON_OPTIONS+=("--cross-file" "$BOB_CWD/cross_file.txt")
                 MESON_OPTIONS+=("-Ddefault_library=${BASEMENT_LIBS:-shared}")
             else
+                MESON_OPTIONS+=("--native-file" "$BOB_CWD/native_file.txt")
```
jkloetzke, line 66: Why is this needed for host builds?

recipes/multimedia/gstreamer.yaml
```diff
@@ -0,0 +1,35 @@
...
+    scm: url
+    url: https://gstreamer.freedesktop.org/src/gstreamer/gstreamer-${PKG_VERSION}.tar.xz
+    digestSHA1: 2a417e7c5eb184b705e0964bb63f83f1498bdf84
+
+buildVars: [PKG_VERSION]
+buildTools: [flex, bison]
+buildScript: |
+   export BISON_PKGDATADIR="${BOB_TOOL_PATHS[bison]}/../share/bison"
+   mesonAddTool bison "${BOB_TOOL_PATHS[bison]}/bison-real"
+   mesonAddTool flex "${BOB_TOOL_PATHS[flex]}/flex"
```
jkloetzke, line 23: That looks odd. According to the documentation tools are searched in `$PATH` automatically. Also, why is it necessary to re-do the bison wrapper script here?

jkloetzke (commented):
Most of the changes left me wondering why they are necessary. I merged the long running PR for the Windows cmake changes that included some tests for meson too (tests/linux). They are working for native and cross compiling so far...

classes/meson.yaml
```diff
@@ -19,30 +19,34 @@ buildScript: |
...
+    fi
 
-        cat >cross_file.txt <<EOF
+    cat >$cross_native <<EOF
     [binaries]
     c     = '$CC'
     cpp   = '$CXX'
     ar    = '$AR'
     strip = '$STRIP'
+    pkgconfig = '${BOB_TOOL_PATHS["pkg-config"]}/pkg-config'
```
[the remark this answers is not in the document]
rhubert: without this meson building gstreamer can't find pkg-config and fails later when looking for `glib`

```
Checking whether type "struct itimerspec" has members "it_interval", "it_value" : YES 
Pkg-config binary for MachineChoice.HOST is not cached.
None of 'PKG_CONFIG' are defined in the environment, not changing global flags.
Pkg-config binary missing from cross or native file, or env var undefined.
Found Pkg-config: NO
Pkg-config binary for machine MachineChoice.HOST not found. Giving up.
[....]
Run-time dependency glib-2.0 found: NO (tried pkgconfig and cmake)
```

[log.tar.gz](https://github.com/BobBuildTool/basement/files/5686747/log.tar.gz)

classes/meson.yaml
```diff
@@ -52,9 +56,14 @@ buildScript: |
         if [[ ! -e .meson-done ]] ; then
             declare -a MESON_OPTIONS=( "$1" "$PWD")
             if [[ -e $BOB_CWD/cross_file.txt ]] ; then
+
+                if [[ "${MESON_FILE:-undefined}" != "undefined" ]]; then
+                    cat ${BOB_TOOL_PATHS['target-toolchain']}/${MESON_FILE} >> ../$cross_native
+                fi
```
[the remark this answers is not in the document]
rhubert: I thought about this but couldn't find a easy way to translate this. (it was easier to add a host-file...) Will look into the translation again. ATM the `cpu` information is not available?

classes/meson.yaml
rhubert: I don't think it's necessary but with this the same file as for cross compiling can be used to tell meson about the system. For me this makes thinks more clear and the class easier to understand..

recipes/multimedia/gstreamer.yaml
rhubert: Without this `gstreamer` build fails cause `bison` and `flex` can't be found. Directly executing the bison wrapper also failed with `exec format error`.

rhubert: would it be possible to extend bob in order to remove the wrapper by using 

```
provideVars:
 BISON_PKGDATADIR: "$SOME_COOL_BOB_VAR/../share/bison"
```

in the `bison.yaml`?

jkloetzke: No, that won't work because all variables are computed independent of the build location.

I'm currently testing your changes. The `exec format error` is caused by a missing `#!/bin/sh` header of the wrapper script. Unfortunately this leads to a rebuild of *everything*. :sob:

jkloetzke:
I've tried to built the recipes here but found numerous issues that surfaced when building in the sandbox. I took the freedom to push the result onto your branch. I have not run any of the results. I did just built them...

rhubert:
Thanks!
But for some reasons the cmake build is now broken :(  There are many undefined references to libssl:

```
x86_64-bob_compat-linux-gnu/1/workspace/usr/lib/libcurl.a(libcurl_la-md5.o): In function `Curl_md5it':
md5.c:(.text+0x17): undefined reference to `MD5_Init'
```

Looks like [this patch](https://github.com/BobBuildTool/basement/blob/master/recipes/devel/cmake/0001-fix-curl-lib-deps.patch) is no longer working? Any ideas?

rhubert:
The pkgconfig files were broken...now cmake builds again. (the remaining build is still pending - rebuilding after changing the install class take some time...)

Issue #40, opened 2020-12-16
forward python for ninja build


mahaase:
Hi.

Ninja needs python in buildScript:

https://github.com/BobBuildTool/basement/blob/a9d3036eae50278b5e5dfad3bbafd00a649e2621/recipes/devel/ninja.yaml#L14

but there is no python tool forwarded in rootrecipe:

https://github.com/BobBuildTool/basement/blob/a9d3036eae50278b5e5dfad3bbafd00a649e2621/classes/basement/rootrecipe.yaml#L44

BR.

jkloetzke:
Did you have a build failure or was it just an observation?

Python is assumed to be available one the host and is also installed in the sandbox image. The `python::python3-minimal` tool is only used during sandbox builds.

mahaase:
yep.

/usr/bin/env python: no such file or directory.

jkloetzke:
With or without sandbox? It could be that `/usr/bin/env` is missing or `python` is not in `$PATH`...

mahaase:
build withour sandbox. no "python", just "python3".

mahaase:
adding python3-minimal in front of ninja failes with that one:
https://stackoverflow.com/questions/59618277/python3-cross-compilation-make-failure 

no idea how to fix that.

jkloetzke:
No, trying to add python3-minmal is not the right approach. Even if you could build meson it might later fail for the same reason when trying to invoke it. Either we call `python3` directly or we require that `python` is available on the host. In the sandbox it is explicitly done this way.

So replacing

    "$1/configure.py" --bootstrap

with

    python3 "$1/configure.py" --bootstrap

might be the simplest solution...

mahaase:
okay. for this use-case i am fine with (BUT i dont understand, why we shouldnt use a self-built python3 host tool). 

i try to port our recipes by using the basement now!

here we need a python3 host module. because we have python generators we provide with bob, so we also want to provide the python3 tool too. for me, it is not possible to build a host python tool with the host-toolchain (because of the issue above).

but this is maybe a different ticket?!

jkloetzke:
I don't quite understand. On the host python3 is already required because you use Bob. In the sandbox python3 is also available. What is missing is the standard "python" executable that *should* point to python3 nowadays but used to be python2 on many systems. Apparently on your distro it is missing altogether. I consider it a problem of the meson package to require an unqualified python version which asks for trouble IMHO.

[...]

PR #43, opened 2020-12-27
.initial commit for .relocatable. python3 in basement

daxcore:
hey. 

this is a prototype of my idea to add python3 to basement: https://github.com/BobBuildTool/basement/issues/40

- relocatable, not 100%, but it does, what it shall
- some other necessary fixes
- a new class for python3 stuff, like a pip install
- added the usage-example to the existing libs: libholler and libgreeter.

surely not perfect, but maybe with power of the community a cool, new feature?!


recipes/python/python3.yaml
```diff
@@ -53,27 +55,40 @@ multiPackage:
...
                 --disable-test-modules \
                 --disable-lib2to3 \
                 --disable-tk \
                 --disable-nis \
                 --disable-idle3 \
-                --disable-pyc-build
+                --disable-bzip2 \
+                --disable-xz \
+                --disable-pyc-build \
+                --disable-uuid
```
daxcore, line 70: or we add e2fsprogs library... if necessary.

recipes/python/python3.yaml
```diff
@@ -53,27 +55,40 @@ multiPackage:
...
 
+        provideTools:
+            python3: usr/bin
+
     minimal:
         depends:
            - libs::expat-dev
+           - libs::gdbm-dev
            - libs::libffi-dev
+           - libs::zlib-dev
```
daxcore, line 91: still needed for gdbm (also if --disable-zlib), which is also a must-have...

tests/cross-platform/recipes/cmake/libholler.yaml
```diff
@@ -7,11 +7,13 @@ checkoutSCM:
...
 
 multiPackage:
     dev:
         packageScript: |
            cmakePackageDev
+           installPythonModule
     tgt:
         packageScript: |
-           cmakePackageTgt
+           cmakePackageTgt "$1" "!/usr/lib/python*"
```
daxcore, line 19: or we have to "fix" the install class. (full rebuild necessary)

tests/cross-platform/recipes/cmake/libgreet.yaml
```diff
@@ -12,6 +12,23 @@ checkoutSCM:
     prune: True
 
 buildScript: |
+   cat <<EOF > myfile.py
```
daxcore, line 15: this could be e.g. a python-code-generator...

classes/python3.yaml
```diff
@@ -0,0 +1,36 @@
+buildTools: [python3]
+buildVars: [PYTHONPATH]
+buildScript: |
+    #if [[ ${BOB_HOST_PLATFORM} == "linux" ]]
+    export PYTHONPATH
+    for i in "${!BOB_DEP_PATHS[@]}" ; do
+        if [[ $i == *-dev ]] ; then
+            for j in $(find ${BOB_DEP_PATHS[$i]}/usr -type d -name "python3*") ; do
```
daxcore, line 8: we can also get the paths by using pythons sysconfig module!

jkloetzke:
Having a relocatable python as tool is definitely worth it. A couple of questions come to my mind:

* Wouldn't that replace `python3-minimal`? This is the current version of a very minimal python3 tool. If we can have a full featured, relocatable one then let's use this. The standard python3 package 
* Does this all still build in the sandbox? The ncurses `--enable-widec` and python3 `--with-ensurepip` look scary.
* The tests should be separate from the cmake tests. It also does not look cross-platform yet. Does this work on MSYS?

daxcore:
in all my tests i needed `--enable-widec`, without it, python build will fail, while linking time:
```
cannot find -lpanelw
cannot find -lncursesw
```

`--with-ensurepip` is one of the main features to install the necessary modules.
but we could use the `get-pip.py` script instead. so we can set a fixed version number (no internal automatic version-changing: necessary for artifacts).

For MSYS2 we have to invest some more, but it should work. we have to use the "embed" version of python3 for windows.
We should not build it with the host-MSYS toolchain. We need a native windows python3 tool, or we will run into trouble with path-stuff later. (The same like, we have to use a native CMake tool, and can't use the MSYS2 CMake tool.)

another thing is the venv module of python3. but i got no good solution to use this instead of the `--root` `--prefix` of `python3 -m pip install`.

daxcore:
> `--with-ensurepip` is one of the main features to install the necessary modules.
> but we could use the `get-pip.py` script instead. so we can set a fixed version number (no internal automatic version-changing: necessary for artifacts).

ohh fail, i see there is already a python3-setuptools recipe... so just adding pip and wheel in the same way and will not need all this stuff above.

daxcore:
update: adapted to MSYS2 (including retest in linux) - works fine for me.

btw: will that be a problem, if python bytecode will be created in package-directories, when the packages will be imported? should we force `pyton3 -B` option?

classes/basement/rootrecipe.yaml
```diff
@@ -65,6 +65,15 @@ depends:
         - name: utils::e2fsprogs
           use: [tools]
           forward: True
+        - name: python::python3-minimal
```
daxcore, line 68: minimal: just python
"": python + setuptools
full: python + setuptools + pip + wheel

classes/python3.yaml
```diff
@@ -0,0 +1,48 @@
...
+    done
+    export PYTHONPATH
+
+    pip_install()
+    {
+        # force lib instead of Lib for MSYS
+        mkdir -p install/usr/lib
+        python3 -m pip install $* \
+            --root ${PWD}/install \
+            --prefix usr \
```
daxcore, line 35: do not use "/usr". this will fail in MSYS, but just "usr" will also work in linux!

classes/python3.yaml
```diff
@@ -0,0 +1,48 @@
...
+            --isolated \
```
daxcore, line 36: maybe not necessary?!

tests/cross-platform/recipes/cmake/greeter.yaml
```diff
@@ -18,11 +21,21 @@ packageScript: |
 
 multiPackage:
     host:
+        environment:
+            HOST: "1"
         depends: &deps
+            - if: !expr |
+                "${HOST}" == "1"
```
daxcore, line 28: how to do that better? xD

daxcore:
added patch for #44

daxcore:
finished. works for me fine. what do u think about this?

daxcore:
so this merge-request is obsolete now!
i restructured all changes in multiple merge-request, with clean commit-messages ;-)

I guess we should close this one here.

Issue #37, opened 2020-11-25
openssl: ${ARCH}-linux will not work for aarch64


mahaase:
Hi.

https://github.com/BobBuildTool/basement/blob/cfc1360a853461b868b8157704225527fe2e6b30/recipes/libs/openssl.yaml#L35

value for ARCH is "arm64" for aarch64 systems.
This will not work with openssl profiles: https://github.com/openssl/openssl/blob/master/Configurations/10-main.conf#L764

We have to use "linux-aarch64".

BR.

jkloetzke:
Yes. Needs to be fixed.

mahaase:
could we also fix, that the stuff will be built statically for host-usage, like default for autotools a.s.o.?!

jkloetzke:
Sure. Patches are welcome. :)

mahaase:
we should wait for openssl 1.1.1i: https://github.com/openssl/openssl/pull/13218
this fix is important to get aarch64 static builds done.
I will create a patch soon.

PR #45, opened 2021-01-05
.update openssl to 1.1.1i and added aarch64 support, fixed static vs.…

daxcore:
fix for #37

### DIFF
--- a/recipes/libs/openssl.yaml
+++ b/recipes/libs/openssl.yaml
@@ -1,7 +1,7 @@
 inherit: [make, install]
 
 metaEnvironment:
-    PKG_VERSION: "1.1.1d"
+    PKG_VERSION: "1.1.1i"
 
 depends:
     - libs::zlib-dev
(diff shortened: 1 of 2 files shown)


recipes/libs/openssl.yaml
```diff
@@ -13,24 +13,68 @@ depends:
...
-    digestSHA256: 1e3a91bc1f9dfce01af26026f856e064eab4c8ee0a8f457b5ae30b40b8b711f2
+    digestSHA256: e8be6a35fe41d10603c3cc635e93289ed00bf34b79671a3a4de64fcee00d5242
     extract: False
 
 buildTools: [target-toolchain]
-buildVars: [PKG_VERSION, CC, AR, RANLIB, ARCH, AUTOCONF_BUILD, AUTOCONF_HOST]
+buildVars: [PKG_VERSION, CC, AR, RANLIB, ARCH, AUTOCONF_BUILD, AUTOCONF_HOST, BASEMENT_LIBS]
 buildScript: |
     export TARGETMACH=${AUTOCONF_HOST}
-    export LDFLAGS="-Wl,-rpath-link=${BOB_DEP_PATHS[libs::zlib-dev]}/usr/lib"
```
daxcore, line 23: the rpath stuff isn't a good idea for being relocatable, is it?

jkloetzke: The rpath-link does not end up in the binary. It's there to find transitive shared library dependencies. The cpackage class normally takes care of this and could indeed be used here. Did you test if it still builds in all cases after removing this line?

https://stackoverflow.com/questions/24598047/why-does-ld-need-rpath-link-when-linking-an-executable-against-a-so-that-needs

recipes/libs/openssl.yaml
```diff
@@ -40,6 +84,10 @@ buildScript: |
 
     popd
 
+    if [[ "$SHARED_ONLY" == "true" ]] ; then
+        find install -iname "*.a" -exec rm -rf {} \;
+    fi
+
```
jkloetzke, line 90: Why is this necessary? The install class takes care of this already. Static libraries are only kept in -dev packages.

daxcore: for both, openssl and zlib, it isn't possible to just build shared libraries. static libraries will be built everytime, and shared libs are optional and additional by flag. but than "both" and "shared" are the same result.

btw. if i choose "both", which one will be used implicitly?

recipes/libs/openssl.yaml
daxcore: Not removed anymore, just added to the configure directly.

jkloetzke:
Could you please structure this as two commits: one for zlib and one for OpenSSL? Please stick to the convention used in the other commits for the commit message.

Have you re-built the sandbox with these changes?

daxcore:
> Could you please structure this as two commits: one for zlib and one for OpenSSL? Please stick to the convention used in the other commits for the commit message.

Okay, of course.

> Have you re-built the sandbox with these changes?

I will check this!

daxcore:
done!?

PR #46, opened 2021-01-20
.update cleanup the vs toolchain path

daxcore:
this could be done a bit easier.


jkloetzke:
Err, no, I'm afraid this is not a good idea. The path will have an impact on the build-id of your artifacts. So you won't be able to share artifacts unless Visual Studio is installed everywhere in the same location. I would also consider it a bug that absolute paths are allowed in `path`. The documentation says

> The path attribute is always needed. The libs attribute, if present, must be a list of paths to needed shared libraries. Any path that is specified must be relative.

So it would not be wise to rely on the glitch...

daxcore:
Ohh, thanks!! Very good to know :-)

PR #47, opened 2021-01-22
devel/win/msbuild: provide the msbuild windows build tool

daxcore:
* added the msbuild tool
* added a class to use the msbuild tool: added the tool to PATH
* get the path by vswhere py plugin


mahaase:
ping, that's up with that one here?
is there still something to do?

Ferruck (commented):
Apart from some seemingly superfluous empty lines in `classes/msbuild.yaml` this looks technically correct to me. I cannot say anything about it functionality wise, though.

jkloetzke:
Thanks! Also for the ping. It just slipped through...

PR #51, opened 2021-01-22
libs/ncurses: enabled widec, added BASEMENT_LIBS (static/shared) switch, small bugfixes

daxcore:
* widec is prerequisite for for other packages, e.g. python3-dev


jkloetzke:
Hmm, we're replicating this case statement over and over again. This should probably be moved into a helper into the cpackage class.

The `--enable-widec` change should also be a separate commit from the shared/static handling. They are independent of each other.

daxcore:
done: restructured commits.

PR #52, opened 2021-01-22
python/python3-modules: added pip + wheel, small fixes

daxcore:
* small fixes for setuptools
* added pip
* added wheel


mahaase:
if not needed, we could close this one. all this is also included in the pipython3 layer repo here: 
https://github.com/mahaase/pipython3

jkloetzke:
I've just merged #166 where I added you as a co-author. Thanks for putting this forward...

PR #53, opened 2021-01-22
python/[win/]python3: relocatable python3 for linux and MSYS2

daxcore:
* depends on #52 
* including test example

### DIFF
--- a/classes/basement/rootrecipe.yaml
+++ b/classes/basement/rootrecipe.yaml
@@ -65,6 +65,15 @@ depends:
         - name: utils::e2fsprogs
           use: [tools]
           forward: True
+        - name: python::python3-minimal
+          use: [tools]
+          forward: True
+        - name: python::python3
+          use: [tools]
+          forward: True
+        - name: python::python3-dev
+          use: [tools]
+          forward: True
 
     - if: !expr |
         "${BOB_HOST_PLATFORM}" == "msys"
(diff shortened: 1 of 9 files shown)


daxcore:
Would it be a better idea to port this into a new basment layer?

jkloetzke:
> Would it be a better idea to port this into a new basment layer?

I'm inclined to agree. As the basement layer is growing I'm more and more open to separate things. In the mid-term the basement layer should be reduced to the core to provide a foundation for all common build systems (autotools, cmake, meson, ...), the C/C++-toolchain and a sandbox. Everything above that should be separated in a "GNU/Linux" layer I guess.

mahaase:
https://github.com/mahaase/pipython3 for https://github.com/BobBuildTool/pipython3.git ? :)

jkloetzke:
This looks very good. I'm just not sure about the naming of the layer. My first thought was `python3-portable`. The provided package could be called `python::portable::python3`, `python3-portable::python3` or just`python3-portable`. :thinking: 

We could also rename the python3 package in basement to `python::basement::python3` to emphasize the fact that this is just a basic python version to build other packages. In the long run the basement layer will certainly be reduced to it's core: providing the basement to build packages with all popular build systems (make, automake, cmake, meson). There will be GNU/Linux layer and other more specialized ones like pipython3. In the end the "standard" GNU/Linux layer will most likely take over the `python::python3` name...

mahaase:
than I would prefer to use `python::portable::python3`. it looks like the most modular one. but i do not have the wide birdview like u. it's ur decission! i will rename it.

daxcore:
Finally that is solved - thanks a lot.

Issue #56, opened 2021-01-27
MSYS2: link tool error (while builing boost library)


daxcore:
https://github.com/BobBuildTool/basement/blob/87de15078a9a3325f8e2e335bd68c17bd97d7a60/classes/cpackage.yaml#L17

in my case the /usr/bin/link tool is picked. we should push `VS_PATH` in front of `PATH`.

mahaase:
in my case i tried to build the boost library. b2 uses the `link` tool.

```
$ which -a link
/usr/bin/link
/bin/link
/c/Program Files/Git/usr/bin/link
```

is there a way to force the correct tool inside the toolchain recipe?

jkloetzke:
Well, I guess there is no good solution. Maybe you have to patch boost and pass an absolute path for the link tool?

mahaase:
perfekt trigger:

```
echo "using msvc : ${MSVC_VER} : ${CXX} : <linker>\"\\\"$(cygpath -m "$(dirname "$(which cl)")/link.exe")\\\"\" ;" > user-config.jam
        EXTRA+=( "--user-config=$PWD/user-config.jam" )
```

take care off the double quote: https://github.com/boostorg/build/issues/566

PR #57, opened 2021-01-27
classes/cpackage: pushed VS_PATH in front of PATH if available

daxcore:
fix for #56


daxcore:
With this "fix", some other (cmake) builds are broken...

Issue #58, opened 2021-01-29
Fails to build


jkloetzke:
readline: https://ci.bobbuildtool.dev/jenkins/view/basement/job/basement-libs__readline-1/25/

```
x86_64-wendy-linux-gnu-gcc -shared -Wl,-soname,libhistory.so.8.0 -Wl,-O1 -Wl,--hash-style=gnu -L/home/jenkins/workspace/basement-libs__readline-1/libs/ncurses-dev/x86_64-wendy-linux-gnu/dist/17-7590-3fdb/usr/lib -L./lib/termcap -Wl,-rpath,/usr/lib -Wl,-soname,`basename libhistory.so.8.0 .0` -o libhistory.so.8.0 history.so histexpand.so histfile.so histsearch.so shell.so mbutil.so xmalloc.so xfree.so -ltinfo
mv compat.o compat.so
mv parse-colors.o parse-colors.so
/home/jenkins/bob/bc/d8bc613012ef866187e74fd2318b08476bc549450a9387162ea23ee9799e2aad55f4a07b4df216-2/content/usr/bin/../lib/gcc/x86_64-wendy-linux-gnu/9.2.0/../../../../x86_64-wendy-linux-gnu/bin/ld: cannot find -ltinfo
collect2: error: ld returned 1 exit status
make[1]: *** [Makefile:181: libhistory.so.8.0] Error 1
make[1]: *** Waiting for unfinished jobs....
make[1]: Leaving directory '/home/jenkins/workspace/basement-libs__readline-1/libs/readline-tgt/x86_64-wendy-linux-gnu/build/17-7590-3fdb/build/shlib'
make: *** [Makefile:219: shared] Error 2
```

Python3: https://ci.bobbuildtool.dev/jenkins/view/basement/job/basement-python__python3-1/26/

```
Failed to build these modules:
_curses               _curses_panel                            

make: *** [Makefile:616: sharedmods] Error 1
```

Most probably regression from #51.

jkloetzke:
@daxcore Could you please have a look at this?

mahaase:
Hey, i detected the problem, there is a patch file for `readline`, with statically searchs for `tinfo` library.
The name of the library with widec feature is suffixed with a 'w' now .

https://github.com/BobBuildTool/basement/blob/7b4528fb883caa9376256da291e6f9201a360ee5/recipes/libs/readline/0000-curses-link.patch#L13

[...]

PR #60, opened 2021-01-29
devel/toolchain: added fPIC option

mahaase:
fix for #58

### DIFF
--- a/recipes/devel/compat/cross-toolchain.yaml
+++ b/recipes/devel/compat/cross-toolchain.yaml
@@ -41,6 +41,6 @@ provideTools:
 
             # compiler flags
             CPPFLAGS:   "-Wdate-time"
-            CFLAGS:     "-O${BASEMENT_OPTIMIZE}$(if-then-else,${BASEMENT_DEBUG}, -g,) -pipe"
-            CXXFLAGS:   "-O${BASEMENT_OPTIMIZE}$(if-then-else,${BASEMENT_DEBUG}, -g,) -pipe"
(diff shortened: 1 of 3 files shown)


jkloetzke:
No, unconditionally adding `-fPIC` is not useful. Every package that builds a shared library should do it by itself. Static libraries do not need it unless you link static libraries into shared libraries. But this is shady to begin with.

In the future we might add `-fPIE` to support position independent executables. But that's a whole different topic.

mahaase:
> Static libraries do not need it unless you link static libraries into shared libraries. But this is shady to begin with.

I have to ask again, why this is shady to link static libraries into shared ones?
Isn't it a part of software-architecture to declare the type of the module?

the current system will handle ALL libraries in the same way by BASEMENT_LIBS.

but some projects define the types of the modules.

module a shall be static
module b shall be shared (maybe because it is part of a delivery)
is there a reason, why that is not correct?

jkloetzke:
It really depends on the particular configuration. As long as the static library is only linked into one shared library and *not* linked to the final executable it should work. Anything else asks for trouble.

Suppose the we have the static libraries S1, S2 and dynamic libraries D1 and D2. The "L1 --> [L2, L3]" notation means L1 links to L2 and L3 resp. L1 depends on L2+L3. The executable is X. The following configurations will exhibit more or less subtle problems:

1. X --> [S1, D1], D1 --> [S1]
2. X --> [D1, D2], D1 --> [S1], D2 --> [S1]
3. X --> [S1, D1], S1 --> [S2], D1 --> [S2]

The most prominent example is double-initialization of global objects. But also duplicated objects are possible. As a general rule a shared library should only depend on other shared libraries. Then you're safe.

PR #61, opened 2021-01-29
libs/ncurses: added fPIC option

mahaase:
fix for #58, if adding the fPIC option to toolchains isn't a good idea.

### DIFF
--- a/recipes/libs/ncurses.yaml
+++ b/recipes/libs/ncurses.yaml
@@ -10,7 +10,7 @@ checkoutSCM:
     stripComponents: 1
 
 buildTools: [host-toolchain]
-buildVars: [BASEMENT_LIBS]
+buildVars: [BASEMENT_LIBS, CFLAGS, CXXFLAGS]
 buildScript: |
     # We need to pass the path of the just built tic
     # able to run on the building machine, so that the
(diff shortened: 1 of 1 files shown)


jkloetzke:
Why is this flag needed?

mahaase:
```
/home/jenkins/bob/bc/d8bc613012ef866187e74fd2318b08476bc549450a9387162ea23ee9799e2aad55f4a07b4df216-2/content/usr/bin/../lib/gcc/x86_64-wendy-linux-gnu/9.2.0/../../../../x86_64-wendy-linux-gnu/bin/ld: /home/jenkins/workspace/basement-python__python3-1/libs/ncurses-dev/x86_64-wendy-linux-gnu/dist/17-7590-3fdb/usr/lib/libncursesw.a(lib_set_term.o): relocation R_X86_64_32S against undefined symbol `_nc_outch_sp' can not be used when making a shared object; recompile with -fPIC
```

So `ncurses` doesn't set the `-fPIC` by itself, but it is necessary. No idea, why this just occurs with my changes... :-/

jkloetzke:
The error was coming from the fact that the ncurses recipe built static libs where shared libs should have been created. No extra flags are needed once this is fixed.

Issue #64, opened 2021-02-04
move gettext to devel and provideTools


mahaase:
https://github.com/BobBuildTool/basement/blob/64767fd751d736087576048dda6a93b3693750e1/recipes/libs/gettext.yaml#L1-L16

i would prefer to move this recipe to devel directory and provide the built tools.

PR #67, opened 2021-02-04
Feature/gettext devel tools

daxcore:
Fix for #64


jkloetzke:
It's also a library but the devel category seems indeed more appropriate.

PR #75, opened 2021-02-19
Feature/graphviz

mahaase:


recipes/libs/cairo.yaml
```diff
@@ -0,0 +1,40 @@
+inherit: [autotools]
+
+metaEnvironment:
+    PKG_VERSION: "1.14.8"
```
Ferruck, line 4: Any reason to use this five year old version? I mean, even if you'd like to stick in the 1.14 line, there's 1.14.12, which is *just* 3 years old...

recipes/libs/fontconfig.yaml
```diff
@@ -0,0 +1,42 @@
+inherit: [autotools, autoconf]
+
+metaEnvironment:
+    PKG_VERSION: "2.13.91"
```
Ferruck, line 4: 2.13.93 is available

recipes/libs/freetype.yaml
```diff
@@ -0,0 +1,29 @@
+inherit: [autotools]
+
+metaEnvironment:
+    PKG_VERSION: "2.10.1"
```
Ferruck, line 4: 2.10.4

recipes/libs/fribidi.yaml
```diff
@@ -0,0 +1,20 @@
+inherit: [autotools]
+
+metaEnvironment:
+    PKG_VERSION: "1.0.5"
```
Ferruck, line 4: 1.0.10

recipes/libs/harfbuzz.yaml
```diff
@@ -0,0 +1,31 @@
+inherit: [autotools]
+
+metaEnvironment:
+    PKG_VERSION: "2.6.0"
```
Ferruck, line 4: 2.7.4

recipes/libs/pango.yaml
```diff
@@ -0,0 +1,40 @@
+inherit: [autotools]
+
+metaEnvironment:
+    PKG_VERSION: "1.40.3"
```
Ferruck, line 4: 1.48.2

recipes/libs/pixman.yaml
```diff
@@ -0,0 +1,20 @@
+inherit: [autotools]
+
+metaEnvironment:
+    PKG_VERSION: "0.38.4"
```
Ferruck, line 4: 0.40.0

Ferruck (commented):
I wonder if it'd be a good idea to switch to SHA-256 sums at least for new recipes. @jkloetzke?

jkloetzke:
> I wonder if it'd be a good idea to switch to SHA-256 sums at least for new recipes. @jkloetzke?

It's probably better. I wouldn't change existing recipes but for new recipes/upgrades I would indeed prefer SHA256.

recipes/libs/fontconfig.yaml
```diff
@@ -0,0 +1,42 @@
...
+
+depends:
+    - name: devel::gettext
+      use: [result, deps, tools]
+    - name: devel::gperf
+      use: [result, deps, tools]
```
jkloetzke, line 10: This won't work in case of cross compiling. The tools won't run on the host.

recipes/libs/pango.yaml
```diff
@@ -0,0 +1,40 @@
...
+
+depends:
+    - libs::glib-dev
+    - name: libs::glib-tools
+      use: [tools]
```
jkloetzke, line 9: ditto. The tool remapping feature could help here, though:

```yaml
tools:
    target-toolchain: host-toolchain
```

jkloetzke (commented):
It's preferred to have the version number in the commit message, e.g. "libs/foobar: add version 1.2.3"

recipes/libs/cairo.yaml
mahaase: i'm stucked ;-)

recipes/libs/fontconfig.yaml
mahaase: falling back to 2.13.92, because 2.13.93 doesn't built here.

recipes/libs/pango.yaml
mahaase: new build-tool. switched to meson.

mahaase:
updated by force-push. i hope i addressed everything (correctly). thx.

recipes/devel/graphviz.yaml
```diff
@@ -0,0 +1,36 @@
...
+      depends:
+          - libs::cairo-tgt
+          - libs::pango-tgt
+          - libs::fontconfig-tgt
+          - libs::expat-tgt
+
+checkoutSCM:
+    scm: url
+    url: https://www2.graphviz.org/Packages/stable/portable_source/graphviz-${PKG_VERSION}.tar.gz
+    digestSHA1: 2cda62953bd84b945c4ae4fe7067da9cb26fc937
```
jkloetzke, line 22: Switch to `digestSHA256`?

recipes/libs/cairo.yaml
```diff
@@ -0,0 +1,40 @@
...
+          - libs::freetype-tgt
+          - libs::fontconfig-tgt
+
+checkoutSCM:
+    scm: url
+    url:  https://www.cairographics.org/releases/cairo-${PKG_VERSION}.tar.xz
+    digestSHA256: 8c90f00c500b2299c0a323dd9beead2a00353752b2092ead558139bd67f7bf16
+    stripComponents: 1
+
+buildVars: [PKG_CONFIG_LIBDIR]
```
jkloetzke, line 27: I don't think this is needed, is it?

recipes/libs/fontconfig.yaml
```diff
@@ -0,0 +1,46 @@
+inherit: [autotools, autoconf]
+
+metaEnvironment:
+    PKG_VERSION: "2.13.92"
+
+depends:
+    - name: devel::gettext
+      use: [result, deps, tools]
```
jkloetzke, line 8: Do you really need the result and the transitive dependencies? That would mean it's still broken because if fontconfig links with a binary of gettext it will have the wrong architecture. Does `use: [tools]` suffice?

jkloetzke (commented):
Looks promising. Only some minor things caught my eye...

recipes/libs/fontconfig.yaml
```diff
@@ -0,0 +1,46 @@
...
+      tools:
+          target-toolchain: host-toolchain
```
mahaase, line 10: i that the way we do it in future? 
when we add tools to the rootrecipe, when we do it like this?

mahaase:
> updated by force-push. i hope i addressed everything (correctly). 

:)

Ferruck (approved):

recipes/libs/fontconfig.yaml
jkloetzke: There is no hard rule but I would put it this way: if a tool is used often, especially in a class, it should be added to the rootrecipe class. On the other end of the spectrum, if a tool is only used for special purposes, then the tool remapping should be used. Of course there is a grey area in between.

jkloetzke:
Thanks!

Issue #74, opened 2021-02-18
glib-dev pkgconfig


mahaase:
glib recipe doesn't work for compat toolchain correctly. glib-tools is there to provide the tools, but for compat a `bindir` is set. so the tools provided by glib-tools package will not selected.

`/bin/bash: line 1: /home/superq/ws/fdt.v2/dev/dist/libs/glib-dev/x86_64-bob_compat-linux-gnu/2/workspace/usr/bin/glib-mkenums: No such file or directory`

compat:

```
bindir=${prefix}/bin
glib_genmarshal=${bindir}/glib-genmarshal
gobject_query=${bindir}/gobject-query
glib_mkenums=${bindir}/glib-mkenums
```

for cross-toolchain, the pkgconfig file looks like this, this would work in combination with glib-tools:

cross:
```
glib_genmarshal=glib-genmarshal
gobject_query=gobject-query
glib_mkenums=glib-mkenums
```

maybe we should not make an explicit package for tools, but install the tools in dev package?!
```patch
mesonPackageDev "$1" "/usr/lib/glib-2.0/***" "/usr/bin/***"
```

or both? :)

jkloetzke:
I'm afraid I do not fully understand the problem. Could you provide some minimal recipe example that show the issue?

mahaase:
yep.

feature-branch for graphviz:
https://github.com/mahaase/basement/pull/new/feature/graphviz

add it e.g. here:

```patch
diff --git a/tests/cross-platform/recipes/cmake/greeter.yaml b/tests/cross-platform/recipes/cmake/greeter.yaml
index 2bf4086..d130d31 100644
--- a/tests/cross-platform/recipes/cmake/greeter.yaml
+++ b/tests/cross-platform/recipes/cmake/greeter.yaml
@@ -1,5 +1,8 @@
 inherit: [ "basement::rootrecipe", cmake ]

+depends:
+    - devel::graphviz
+
 checkoutSCM:
     scm: import
     url: src/greeter/
```

`bob dev cmake::greeter-cross/ -vv`

error:

```
+ touch .configure.stamp
+ makeParallel
+ [[ -n '' ]]
++ nproc
+ make -j 24
make  all-recursive
make[1]: Entering directory '/home/superq/ws/basement/tests/cross-platform/dev/build/libs/pango-dev/x86_64-bob_compat-linux-gnu/1/workspace/build'
Making all in pango
[...]

PR #77, opened 2021-02-23
libs/glib: remove bindir prefix in pc files

mahaase:
to resolve tools like glib-mkenums by PATH

fix for #74


Ferruck (approved):

jkloetzke:
Thanks. BTW, if you include a "Fixes #..." at the end of your commit message then GitHub will automatically close the ticket once its merged...

PR #82, opened 2021-03-23
devel/graphviz: fixed for sandbox builds

mahaase:


recipes/devel/graphviz.yaml
```diff
@@ -23,6 +23,12 @@ checkoutSCM:
     stripComponents: 1
 
 buildScript: |
+    if [[ "$(which ps2pdf)" == "" ]] ; then
+        export ac_cv_prog_PS2PDF="true"
```
jkloetzke, line 27: Shouldn't we set this unconditionally to `true`? Otherwise the result is not reproducible because it depends on the environment if ps2pdf was available or not...

mahaase: indeed. makes sense. fixed.

PR #84, opened 2021-03-29
default.yaml: use linker flag --hash-style just if we are in a gnu toolchain

### LINKED ISSUE #83: MSYS: unrecognized option '--hash-style=gnu'
First things first: Thanks for the fix: https://github.com/BobBuildTool/bob/issues/395 👍 

now there is a problem with the basement: `unrecognized option '--hash-style=gnu'`

https://github.com/BobBuildTool/basement/blob/52ea909a6cdb2000bf4ea72728667f592da5c582/default.yaml#L43

```
CMake Error at C:/msys64/home/mhaase/ws/fdt.bobrecipes/dev/dist/devel/win/cmake/1/workspace/usr/share/cmake-3.17/Modules/CMakeTestCCompiler.cmake:60 (message):
  The C compiler

    "C:/msys64/usr/bin/gcc.exe"

  is not able to compile a simple test program.

  It fails with the following output:

    Change Dir: C:/msys64/home/mhaase/ws/fdt.bobrecipes/dev/build/devel/parasoft-cpptest/x86_64-pc-msys/1/workspace/build/CMakeFiles/CMakeTmp

    Run Build Command(s):C:/msys64/home/mhaase/ws/fdt.bobrecipes/dev/dist/devel/win/ninja/1/workspace/ninja.exe cmTC_a1873 && [1/2] Building C object CMakeFiles/cmTC_a1873.dir/testCCompiler.c.obj
    [2/2] Linking C executable cmTC_a1873.exe
    FAILED: cmTC_a1873.exe
    cmd.exe /C "cd . && C:\msys64\usr\bin\gcc.exe -Os -pipe -fPIC -DCPPTEST_SHARED=0 -DCPPTEST_NO_THREADS -DCPPTEST_FILE_COMMUNICATION  -Wl,-O1 -Wl,--hash-style=gnu CMakeFiles/cmTC_a1873.dir/testCCompiler.c.obj  -o cmTC_a1873.exe -Wl,--out-implib,libcmTC_a1873.dll.a -Wl,--major-image-version,0,--minor-image-version,0   && cd ."
    /usr/lib/gcc/x86_64-pc-msys/10.2.0/../../../../x86_64-pc-msys/bin/ld: unrecognized option '--hash-style=gnu'
    /usr/lib/gcc/x86_64-pc-msys/10.2.0/../../../../x86_64-pc-msys/bin/ld: use the --help option for usage information
    collect2: error: ld returned 1 exit status
    ninja: build stopped: subcommand failed.
```

mahaase:
.FIXES #83

### DIFF
--- a/default.yaml
+++ b/default.yaml
@@ -40,7 +40,7 @@ environment:
     # position independent code in case static and dynamic libraries are mixed.
     CFLAGS:     "-Os -pipe -fPIC"
     CXXFLAGS:   "-Os -pipe -fPIC"
-    LDFLAGS:    "-Wl,-O1 -Wl,--hash-style=gnu"
+    LDFLAGS:    "-Wl,-O1"
 
 archive:
     backend: http
--- a/recipes/devel/host-compat-toolchain.yaml
+++ b/recipes/devel/host-compat-toolchain.yaml
@@ -87,3 +87,4 @@ provideVars:
     # target-toolchain. ARCH is unchaged, though.
     AUTOCONF_BUILD: "$AUTOCONF_TARGET"
     AUTOCONF_HOST:  "$AUTOCONF_TARGET"
+    LDFLAGS: "$LDFLAGS  -Wl,--hash-style=gnu"


jkloetzke:
I think the proper test is to compare `BOB_HOST_PLATFORM` to `linux`. This option is just an optimization and I'm not sure if it's even needed anymore nowadays.

mahaase:
thx for review. reworked.

jkloetzke:
Except for the commit message (No ".FIXES"; The "Fixes #...." should be on a separate line) its ok.

But I've looked deeper into it and it looks like the optimized linker hash is still not the default. Could you please add a 2nd commit that adds `--with-linker-hash-style=gnu` to `buildGcc()` in `recipes/devel/gcc.yaml` and see if this still builds? That should make the gnu hash style the default and we can safely remove the switch form CFLAGS.

mahaase:
now the difference is, that the optimization (maybe) isn't there, if a build with host-gcc will be started..
fine for me!

jkloetzke:
Oh, good point. I don't think that gcc 5.4 does support the configure switch yet so we probably cannot add `--with-linker-hash-style=gnu` in recipes/devel/compat/gcc.yaml. We could amend `CFLAGS` via `provideVars`, though.

BTW, now that we've switched the default hash style we can drop the remaining `-Wl,--hash-style=gnu` in the various recipes too...

mahaase:
soooo, couldn't we safely remove the default.yaml's `linker-hash-style` stuff, because
```
recipes/devel/compat/cross-toolchain.yaml
recipes/devel/cross-toolchain.yaml
recipes/devel/gcc.yaml
```
already provide the `LDFLAGS`? Maybe the `--with-linker-hash-style=gnu` isn't necessary.
The only thing, which is changed is, if somebody uses basement without the provilded toolchains, than the gcc-gnu-linker-hash-style isn't set be default.yaml. a good thing, if the user doesn't uses gcc ;-)

mahaase:
in `devel/compat/gcc` the `target-toolchain` doesn't provide environment at all. should we add this like in `devel/gcc` done?!

jkloetzke:
Yes, we can remove the `--hash-style=` switch from all recipes. As replacement we must:

* add `--with-linker-hash-style=gnu` to `buildGcc()` in `recipes/devel/gcc.yaml`. This will make it the default for all cross-compiling toolchains.
* add the following at the end in `recipes/devel/host-compat-toolchain.yaml` to set the flag *if* the host compat toolchain is used:
```yaml
    LDFLAGS: "$LDFLAGS  -Wl,--hash-style=gnu"
```
If the host compat toolchain is not used (`BASEMENT_HOST_COMPAT_TOOLCHAIN`== 0) then it's up to the project to either set the flag manually or build without the optimization (which is not a real problem).

Your change to `recipes/devel/compat/gcc.yaml` should be dropped. It only influences the build of `libs::compat::glibc-dev` which is not relevant here.

mahaase:
but this one is finished/fixed, isn't it?
from my side it is ready for reintegration.

mahaase:
ping.

PR #85, opened 2021-03-30
cmake: autoconf host regex for windows is *-pc-* instead of *-win32

mahaase:
…so we also match for MSYS host environment

### DIFF
--- a/classes/cmake.yaml
+++ b/classes/cmake.yaml
@@ -8,14 +8,14 @@ buildScript: |
     for i in "${@:2}" ; do
         # If we build for native Windows a native CMake is used and we need to
         # convert to Windows paths.
-        [[ "$AUTOCONF_HOST" == *-win32 ]] && i="$(cygpath -m $i)"
+        [[ "$AUTOCONF_BUILD" == *-msys ]] && i="$(cygpath -m $i)"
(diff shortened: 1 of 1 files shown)


jkloetzke:
Except from the fact that *-pc-* is way too generic I'm wondering why this is necessary. I guess you are compiling on MSYS for MSYS...

daxcore:
`x86_64-pc-msys` is the string for native msys. vs2019 toolchain resets `AUTOCONF_HOST` than to `x86_64-pc-win32`. in both cases we got windows results, but paths must be MSYS compatible.

jkloetzke:
Now I see. In both cases we use a native win32 cmake. Not sure if this works to build msys executables. Have you tried?

The `-pc-` is the vendor field and almost arbitrary. See https://wiki.osdev.org/Target_Triplet

daxcore:
> Now I see. In both cases we use a native win32 cmake. Not sure if this works to build msys executables. Have you tried?

Yep, worked in my cases.

jkloetzke:
Ok. Then we should test for `AUTOCONF_BUILD` being msys, right? Because we always use a native Cmake...

daxcore:
indeed, would make sense!

mahaase:
please recheck?!

PR #86, opened 2021-05-19
devel/win/graphviz: forced extraction of broken archive

mahaase:
FIXES #414 (bob issue tracker)


jkloetzke:
I took myself the freedom to rephrase the commit message. Otherwise we'll certainly be wondering *why* this was necessary two years down the road...

PR #88, opened 2021-05-20
recipes: mirror variable for some common sources urls

mahaase:
with that it is possible to override the mirror variables in a upstream default.yaml.
we use this to mirror the archives on local servers.

### DIFF
--- a/recipes/devel/flex.yaml
+++ b/recipes/devel/flex.yaml
@@ -5,7 +5,7 @@ metaEnvironment:
 
 checkoutSCM:
     scm: url
-    url: https://sourceforge.net/projects/flex/files/flex-${PKG_VERSION}.tar.gz
+    url: ${SOURCEFORGE_MIRROR}/flex/flex-${PKG_VERSION}.tar.gz
     digestSHA1: "cfe10b5de4893ced356adc437e78018e715818c3"
     stripComponents: 1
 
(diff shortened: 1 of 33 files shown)


jkloetzke:
Look good to me. I think the proper solution is to implement mirror support in Bob directly (BobBuildTool/bob#360). But in the meantime this is certainly ok.

recipes/devel/flex.yaml
```diff
@@ -5,8 +5,8 @@ metaEnvironment:
 
 checkoutSCM:
     scm: url
-    url: https://sourceforge.net/projects/flex/files/flex-${PKG_VERSION}.tar.gz
-    digestSHA1: "cfe10b5de4893ced356adc437e78018e715818c3"
+    url: ${SOURCEFORGE_MIRROR}/flex/flex-${PKG_VERSION}.tar.gz
+    digestSHA256: fc367e9deb570ee7fc7893ce5066c2dbdedf3cf27e93f2f39d1318703b028f0c
```
jkloetzke, line 9: Please do not change the digest algorithm of packages in this PR. It will rebuild major parts for no good reason. We can still switch to SHA256 whenever the package is updated.

recipes/libs/openssl.yaml
```diff
@@ -12,8 +12,8 @@ depends:
 
 checkoutSCM:
     scm: url
-    url: https://www.openssl.org/source/openssl-${PKG_VERSION}.tar.gz
-    digestSHA256: e8be6a35fe41d10603c3cc635e93289ed00bf34b79671a3a4de64fcee00d5242
```
jkloetzke, line 16: Why change to another source?

daxcore:
done. pls recheck.

jkloetzke:
Thanks :)

PR #94, opened 2021-08-23
plugins/path: short-path feature for bob workspace

mahaase:
if u like, feel free to merge this into basement, if not, just close the pull request, so we have this post for documentation reasons :)
---

because of a 250 char path length restriction of windows, it is necessary to keep paths as short as possible

usage:
 - use the MSYS2 paths, e.g. /d/.bob, not windows native like D:\
 - if the shortpath isn't used, the behavior shall be unchanged

 - use a config.cfg file and bob -c config
 - use the user.yaml or not recommended, the default.yaml

 - content of the settings-file:
```
ShortPath:
    path: "/d/.bob"
```

### DIFF
--- a/plugins/paths.py
+++ b/plugins/paths.py
@@ -1,6 +1,21 @@
+from os import makedirs
 from os.path import join
 from bob.errors import ParseError
-from bob.input import PluginState, PluginProperty
+from bob.input import PluginState, PluginProperty, PluginSetting
+
+class ShortPathSetting(PluginSetting):
+    pass
+
+shortPathSetting = ShortPathSetting(None)
+
+def shortPathFormatter(step, states):
+    settings = shortPathSetting.getSettings()
+    shortPath = settings.get('path') if settings else None
+    if shortPath:
+        makedirs(shortPath, exist_ok=True)
+        return shortPath
+    else:
+        return ""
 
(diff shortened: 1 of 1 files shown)


mahaase:
cleaned up. no strange magic inside anymore :)

jkloetzke:
Actually now that I look at this I have the feeling a proper solution would be to support out-of-source builds in Bob. Like:

```bash
mkdir /d/bob/b1 && cd /d/bob/b1
bob init /path/to/recipes
bob dev ...
```

or

```bash
bob init /path/to/recipes /d/bob/b1
cd /d/bob/b1
bob build ...
```

That would keep the paths short and there could be multiple builds of the same recipes (e.g. with different options). Most importantly it prevents any accidental collisions if you configured two projects for the same `ShortPath`...

mahaase:
sounds very good. in our current solution there are some problems, e.g. if i set the path in user.yaml.
on deleting the user-yaml the path will not be resetted to the default path.
so the current solution is not perfect.

jkloetzke:
Thanks for the feedback. I created a tracking ticket for the feature: BobBuildTool/bob#435

I'll close this because it would only be a stop-gap solution that with the drawbacks you mentioned above. Thanks for the effort nonetheless.

PR #98, opened 2021-11-05
libs[/compat]/isl: switch mirror to sourceforge

### LINKED ISSUE #97: isl.gforge.inria.fr is down
https://github.com/BobBuildTool/basement/blob/ad124065ccf6adb5b3118f638b12e1a5cc3b2af4/recipes/libs/isl.yaml#L20

isl.gforge.inria.fr is down. maybe we should switch to sourceforge.

mahaase:
reason: isl.gforge.inria.fr is down

.FIXES #97

### DIFF
--- a/recipes/libs/compat/isl.yaml
+++ b/recipes/libs/compat/isl.yaml
@@ -15,7 +15,7 @@ depends:
 
 checkoutSCM:
     scm: url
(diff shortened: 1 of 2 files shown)


jkloetzke:
Thanks :)

PR #100, opened 2021-11-09
classes/sandbox-toolchain: fixed issue added by #96

mahaase:
changing from packageScript to packageSetup here isn't correct and breaks the sandbox


mahaase:
sorry for that. now the jenkins stuff is running again and I will test all the stuff more intensive :)

PR #101, opened 2022-01-17
libs/glib: Package libraries alongside tools

Ferruck:
At least some of the glib tools such as the glib-resource-compiler need libgio.so in some form or another.

I'm still a bit unsure about the usage of `mesonPackageTgt` here, though: IMHO tools needing shared objects of their own projects shouldn't be *that* rare, so `mesonPackageBin` might actually handle that?!

### DIFF
--- a/recipes/libs/glib.yaml
+++ b/recipes/libs/glib.yaml
@@ -38,6 +38,8 @@ multiPackage:
         packageScript: mesonPackageLib
         provideDeps: [ "*-tgt" ]
     tools:
-        packageScript: mesonPackageBin
+        packageScript: mesonPackageTgt
         provideTools:
-            glib: usr/bin
+            glib:
+                path: usr/bin
+                libs: [ "usr/lib" ]


jkloetzke:
Actually the idea was that host tools do not use shared libraries at all. Maybe this is not working correctly for meson packages. At least the autotools class does take care of this. Shared libraries for tools will get nasty if the library is from a dependency rather than the package itself.

> I'm still a bit unsure about the usage of `mesonPackageTgt` here, though: IMHO tools needing shared objects of their own projects shouldn't be _that_ rare, so `mesonPackageBin` might actually handle that?!

No, `mesonPackageBin` (like all other `*Bin` variants) should only pick up the executable(s). The `*Tgt` variant is the one that installs executables and shared libraries. See the comments in the `install` class...

Ferruck:
> Actually the idea was that host tools do not use shared libraries at all. Maybe this is not working correctly for meson packages. At least the autotools class does take care of this. Shared libraries for tools will get nasty if the library is from a dependency rather than the package itself.

Yes, been there, done that. I'll have a look at the meson class instead, then. I will come back when I have something to share.

Ferruck:
Just as a follow up: The meson class is fine. I was just including my cross toolchain *before* depending on my tools which led to re-usage of the build step of the regular `glib-tgt` package, which is obviously dynamically linked.

PR #102, opened 2022-01-18
classes/cmake: Set CMAKE_SYSTEM_PROCESSOR

Ferruck:
Some CMake scripts access this variable unconditionally, such as [`libjpeg-turbo`'s](https://github.com/libjpeg-turbo/libjpeg-turbo/blob/main/CMakeLists.txt#L57). Looking at the [official docs](https://cmake.org/cmake/help/latest/variable/CMAKE_SYSTEM_PROCESSOR.html) they seem to be right in the assumption that `CMAKE_SYSTEM_PROCESSOR` will always be set.

Content-wise the variable seems to match the first part of the `AUTOCONF_HOST` triplet for all Unix-ish OSs. I'm not sure how the whole cygwin/mingw-stuff has to be handled here, though, since normally Windows [has its own possible settings](https://docs.microsoft.com/en-us/windows/win32/winprog64/wow64-implementation-details?redirectedfrom=MSDN) for `CMAKE_SYSTEM_PROCESSOR` like `AMD64`, `ARM64` or `IA64`. Do we need to take those when compiling via cygwin/mingw or just the same first part of the triplet? I'd say that it'll probably make no difference in the majority of cases but I'd rather provide a correct solution, anyways.

### DIFF
--- a/recipes/devel/doxygen.yaml
+++ b/recipes/devel/doxygen.yaml
@@ -9,7 +9,8 @@ checkoutSCM:
     digestSHA256: 67aeae1be4e1565519898f46f1f7092f1973cce8a767e93101ee0111717091d1
     stripComponents: 1
 
-buildTools: [flex, bison, dot, python3]
+buildTools: [flex, bison, dot]
+buildToolsWeak: [python3]
 buildScript: |
     cmakeBuild "$1"
 


jkloetzke:
Oh well, looks like another loosely defined variable in build systems. I think your approach is good enough for the time being. It might also work for Windows/MSYS2 because any package using this variable probably needs to handle the Linux cases too.

@mahaase Do you use this class on MSYS2?

mahaase:
> @mahaase Do you use this class on MSYS2?

yep, currently result is: `set(CMAKE_SYSTEM_PROCESSOR x86_64)` which matches to https://github.com/BobBuildTool/basement/blob/master/plugins/multiarch.py, doesn't it?

Ferruck:
So, I've made some changes taking into account what I found in the `devel::win::vs2019-toolchain` recipe. I don't know what I'm doing here, though (the linked issue just made the whole mess more confusing, IMHO), and am thus open for comments. I also have no way of testing this locally, so, @mahaase, it'd be nice if you could give it a shot...

Ferruck:
I can just agree with [this comment](https://github.com/conan-io/conan/pull/8026#discussion_r530631240):

> Otoh, I searched for CMAKE_SYSTEM_PROCESSOR arm and CMAKE_SYSTEM_PROCESSOR aarch on github and it looks like there is no universal standard.
> I think we have to face the reality that there is no universally-agreed-to-standard to use as CMAKE_SYSTEM_PROCESSOR variable.
> Like everything in the c/c++ world (except for the stl standaard), there is no standard. CMake allows everything. CMAKE_SYSTEM_PROCESSOR is just a hint.
>
> In reality, some project will either use one kind CMAKE_SYSTEM_PROCESSOR mapping from processor to cmake variable and not switch between a Microsoft convention or GNU convention. Or will not switch between Windows convention or Linux convention.

jkloetzke:
I guess this is as good as it can get. As it's only loosely defined we shouldn't put too much effort into this until a problem is found. Unless @mahaase has some objection I'm going to merge it...

Ferruck:
Sorry for the noise, seems I'm too dumb for that whole Github flow. I've created a new PR (this time not referencing `master` :roll_eyes: ) in #111

PR #105, opened 2022-01-21
libs/glib: Update to 2.71

Ferruck:
The patch is no longer needed since a similar change has been made upstream.


jkloetzke:
Thanks.

PR #108, opened 2022-01-21
libs/fribidi: Update to 1.0.11

Ferruck:


recipes/libs/fribidi.yaml
```diff
@@ -1,20 +1,21 @@
...
+    PKG_VERSION: "1.0.11"
 
 checkoutSCM:
     scm: url
     url: ${GITHUB_MIRROR}/fribidi/fribidi/releases/download/v${PKG_VERSION}/fribidi-${PKG_VERSION}.tar.xz
-    digestSHA256: 7f1c687c7831499bcacae5e8675945a39bacbad16ecaa945e9454a32df653c01
+    digestSHA256: 30f93e9c63ee627d1a2cedcf59ac34d45bf30240982f99e44c6e015466b4e73d
     stripComponents: 1
 
+buildTools: [host-toolchain]
```
Ferruck, line 12: I'm unsure about this one and there's probably a better solution to it, but without this line configure fails with

```
configure: error: cannot find native C compiler
```

daxcore: this happens if configure searchs explicitly for "gcc" or "g++".
normally the target-toolchain compiler is named like: x86_64-cross-linux-gnu-gcc...

Ferruck: The target toolchain is found, actually, but the package needs the host toolchain, too (for tools or tests or IDK)...

Ferruck: Any further input on this? I actually have a second package (`libs::freetype`) needing the exact same "fix" and I'd like to bring my fork back on par with upstream...

jkloetzke: Well if the package needs the host toolchain then I see no problem in using it. Otherwise we would need to patch the package but I don't see a benefit in doing that...

jkloetzke:
Any reason why it should not be merged? LGTM...

mahaase:
ship it! :)

Ferruck:
> Any reason why it should not be merged? LGTM...

Just that it felt fishy with the host toolchain...

PR #111, opened 2022-01-27
classes/cmake: Set CMAKE_SYSTEM_PROCESSOR

Ferruck:
Some CMake scripts access this variable unconditionally, such as
libjpeg-turbo's. Looking at the official docs [1] they seem to be right
in the assumption that CMAKE_SYSTEM_PROCESSOR will always be set.

[1] https://cmake.org/cmake/help/latest/variable/CMAKE_SYSTEM_PROCESSOR.html


jkloetzke:
Thanks!

PR #116, opened 2022-03-04
libs/fontconfig: removed unnecessary autoconf reconfigre

mahaase:
that indroduced some indeterminism in checkout step. so artifacts built e.g. on jenkins in sandbox didn't matched to somewhere else.

### DIFF
--- a/recipes/libs/fontconfig.yaml
+++ b/recipes/libs/fontconfig.yaml
@@ -1,4 +1,4 @@
-inherit: [autotools, autoconf]
+inherit: [autotools]
 
 metaEnvironment:
     PKG_VERSION: "2.13.92"
@@ -26,10 +26,6 @@ checkoutSCM:
     url: https://www.freedesktop.org/software/fontconfig/release/fontconfig-${PKG_VERSION}.tar.xz
     digestSHA256: 506e61283878c1726550bc94f2af26168f1e9f2106eac77eaaf0b2cdfad66e4e
     stripComponents: 1
-
-checkoutDeterministic: True
-checkoutScript: |
-    autoconfReconfigure
 
 buildTools: [gettext, gperf]
 buildScript: |


jkloetzke:
Just out of curiosity: have you tried to find out what the `autoconfReconfigure` left behind that made it indeterministic (e.g. do --checkout-only in two different workspaces)? Because it should not have happened in the first place, even if the reconfigure would be necessary...

mahaase:
i analyzed it and got the following findings `find . -exec sha256sum {} \; | sort`:
```diff
-44c56e9934f233c926719575553ba2ef45d5d3f02e54b6190f64cb498200c60c  ./src/Makefile.in
+64e2c716ecce4988aeed5c4f496dee6b7c3ccd49c4e400b6b6b4d3d20a8026e6  ./src/Makefile.in
-7790292709512094e69d50ed992d9729505aa16adaeb4728836070960bb58c2c  ./fc-cache/Makefile.in
+a38f331f0f1d7fada7c79242375fe6fb1d2796dc8e079fcf0adf67ea956786ef  ./fc-cache/Makefile.in
```
./src/Makefile.in:
```diff
-@ENABLE_SHARED_FALSE@uninstall-local:
-@ENABLE_SHARED_FALSE@install-data-local:
+@ENABLE_SHARED_FALSE@install-data-local:
+@ENABLE_SHARED_FALSE@uninstall-local:
```
./fc-cache/Makefile.in
```diff
-@OS_WIN32_TRUE@uninstall-local:
-@OS_WIN32_TRUE@install-data-local:
+@OS_WIN32_TRUE@install-data-local:
+@OS_WIN32_TRUE@uninstall-local:
```
uhh my gosh. :/

jkloetzke:
This is some indeterminism in automake. Even happens on automake 1.16.3. After one f****g hour I finally found the spot in this gigantic perl script. I'll add a patch soon...

PR #117, opened 2022-03-07
devel/automake: bump to 1.16.5

jkloetzke:
Fixes indeterministic automake outputs that were generated in case of
conditional rules. See #116.


jkloetzke:
The problem was already fixed upstream.

@mahaase @Ferruck Does this break anything for you? I've only built the sandbox so far...

mahaase:
fine for me. 👍

PR #121, opened 2022-03-17
classes/install: fixed typo, replaced stripFile by stripBinary

mahaase:
typo? stripFile doesn't exists.


jkloetzke:
Indeed, looks like a typo. Thanks. :+1:

PR #124, opened 2022-04-02
WIP: devel/compat/gcc: use c++11 standard

daxcore:
necessary for newer gcc versions, like gcc 11.2.0
to prevent the compiler error:
`use of an operand of type 'bool' in 'operator++' is forbidden in C++17`


daxcore:
didn't fixed it.

Issue #123, opened 2022-04-02
compat gcc build error with gcc11 C++17


daxcore:
https://github.com/BobBuildTool/basement/blob/225b0b5e53f609f50345347e307b82eff10cb3a8/recipes/devel/compat/gcc.yaml#L83

it isn't possible to build gcc compat 5.4 with gcc 11.2:

`use of an operand of type 'bool' in 'operator++' is forbidden in C++17`

the internet said, gcc5.4 must be built by c99 and c++03. but no success for me by setting this via C(XX)FLAGS.

daxcore:
ohh, I guess I got it by just setting the `CXXFLAGS` c++ standard < c++17. PR is on the way.

daxcore:
created a new pull request.. this one works for the full chain of dependencies.

jkloetzke:
Interesting. On what system and how specifically was it built? Could you provide the output of `gcc --version -v`?

BTW, we could update the compat gcc to version 7.4. This is the version of the oldest Ubuntu LTS that is still supported (bionic). Maybe that will also solve the problem...

Ferruck:
> Interesting. On what system and how specifically was it built? Could you provide the output of `gcc --version -v`?

Sure:

```
Es werden eingebaute Spezifikationen verwendet.
COLLECT_AS_OPTIONS='--version'
COLLECT_GCC=gcc
COLLECT_LTO_WRAPPER=/usr/lib/gcc/x86_64-pc-linux-gnu/11.2.0/lto-wrapper
gcc (GCC) 11.2.0
Copyright (C) 2021 Free Software Foundation, Inc.
Dies ist freie Software; die Kopierbedingungen stehen in den Quellen. Es
gibt KEINE Garantie; auch nicht für MARKTGÄNGIGKEIT oder FÜR SPEZIELLE ZWECKE.


Ziel: x86_64-pc-linux-gnu
[...]

PR #128, opened 2022-04-08
Compat toolchain upgrade

Ferruck:
This pulls all parts of the compat toolchain up to the versions used in the last supported Ubuntu LTS 18.04.6. The version of the GCC used previously was too old to compile recent U-Boot tools. Additionally, this would probably be a fix for #123 (haven't tested that, @mahaase could you give it a spin?).


Ferruck:
I'm still testing these changes with our AppImage recipes but I expect no problems. I'll report back.

recipes/libs/compat/glibc.yaml
```diff
@@ -32,7 +32,7 @@ buildScript: |
             --prefix=/usr                    \
             --host=${AUTOCONF_TARGET}        \
             --build=$($1/scripts/config.guess) \
-            --enable-kernel=4.3.3             \
+            --enable-kernel=5.4.157          \
```
jkloetzke, line 35: AFAICT Ubuntu 18.04 is on 4.15.0...

daxcore:
no c++17 issues! so i am fine with that solution too!

recipes/libs/compat/glibc.yaml
mahaase: confirm.

Ferruck:
I can confirm our AppImage building and working as before with these changes.

jkloetzke:
Thanks!

PR #131, opened 2022-04-21
Move sub-layers into separate repositories

jkloetzke:
The gnu-linux and win32-bin sub-layers were moved into separate
repositores. Drop them from here so that the basement layer is
restricted to the sandbox, toolchains and support for the most common
build systems.


mahaase:
what's about the default.yaml that isn't available in the new layers?
e.g. the `*_MIRROR` variables still missing. maybe something else too?

jkloetzke:
No, as written in the [README](https://github.com/BobBuildTool/basement-gnu-linux/blob/master/README.md#how-to-use) the separated layers are intended to be used together with the basement layer...

mahaase:
how this looks if both new layers will be used?

```
layers:
- basement-gnu-linux
    - basement
- basement-win32-bin
    - basement
```

btw: the README file of [basement-win32-bin](https://github.com/BobBuildTool/basement-win32-bin) has no information.

jkloetzke:
Ah, I see the indentation in the README is broken. Sorry.

So in your project you do:

```
layers:
    - basement-gnu-linux
    - basement-win32-bin
    - basement
```

Just stack them on top of each other.

mahaase:
so just the information in win32 basement is missing, that it depends on basement.

jkloetzke:
Yes. I thought that it is independent but overlooked the mirrors variables.

mahaase:
i tested it in our company. everything is fine. cool stuff. thx.

PR #132, opened 2022-04-27
libs/libcap: Provide host toolchain

Ferruck:
This was needed for me in order to get cross-compilation to work, otherwise it would fail with an error message about an exec format error in one of the tools used while building.

Also, upgrade to 2.64 whilst we're at it.


jkloetzke:
Thanks.

PR #139, opened 2022-05-20
classes/{autotools(-*)?,meson}: Remove install dir

### LINKED ISSUE #138: CMake class removes "install" directory
[Here](https://github.com/BobBuildTool/basement/blob/b0776cdb84327ecc3d5dff8f33ca64aab5a7171b/classes/cmake.yaml#L138) the CMake class removes the `install` directory before creating a new one and building. The Autotools class does nothing alike. Is either one of these an oversight or is there a reasoning behind that?

The reason I'm asking is that I myself have a class that creates files in the `install` directory which are then removed by the CMake class. I'd rather avoid putting this into a function that I have to remember to call in each recipe after building, if possible.

Ferruck:
Analogous to what the CMake class already does, the install directory
should be removed before each new compilation to avoid conflicts with
old files still lingering around.

I've tested these changes by building our current project. From what I
can tell, the output is the same.

Fixes #138

### DIFF
--- a/classes/autotools-noarch.yaml
+++ b/classes/autotools-noarch.yaml
@@ -37,8 +37,10 @@ buildSetup: |
         done
         shift $(( OPTIND -1 ))
 
+        rm -rf install
         mkdir -p build install
         pushd build
+
         if [[ $1/configure -nt .configure.stamp ]] ; then
             $1/configure \
                 --prefix="/usr" \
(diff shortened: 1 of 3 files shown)


jkloetzke:
I cancelled the CI build because I don't think it will bring up anything sensible. Will merge together with #140 after the CI build succeeded there...

PR #145, opened 2022-06-03
libs/libcap: added support for cross-compiling

### LINKED ISSUE #144: libs/libcap fails at cross-compiling
if cross-compiling we have to set the `BUILD_CC` variable too.

mahaase:
.FIXES #144

### DIFF
--- a/recipes/libs/libcap.yaml
+++ b/recipes/libs/libcap.yaml
@@ -10,7 +10,7 @@ checkoutSCM:
     extract: False
 
 buildTools: [host-toolchain, target-toolchain]
-buildVars: [AR, CC, OBJCOPY, RANLIB, PKG_VERSION]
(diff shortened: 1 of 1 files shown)


recipes/libs/libcap.yaml
```diff
@@ -21,6 +21,7 @@ buildScript: |
         AR=$AR \
         CC=$CC \
         OBJCOPY=$OBJCOPY \
+        ${CROSS_COMPILE:+BUILD_CC=gcc} \
```
jkloetzke, line 24: This line is unnecessary because `BUILD_CC` is always set to `gcc` on line 20 above...

mahaase: hmm @Ferruck was faster than me. 2 month ago, i didn't realized that! so this is fixed already 👍

PR #147, opened 2022-06-03
devel/git: add cross-compilation support

### LINKED ISSUE #142: add cross-compilation support for git
https://github.com/BobBuildTool/basement/blob/cbd2542abc66ba3df1636fb560681be1ab1623d2/recipes/devel/git.yaml#L24-L38

currently it is not possible to cross-compile git, because of missing `--build` argument, we should add this.

mahaase:
.FIXES #142

### DIFF
--- a/recipes/devel/git.yaml
+++ b/recipes/devel/git.yaml
@@ -21,9 +21,17 @@ checkoutSCM:
     digestSHA256: "405d4a0ff6e818d1f12b3e92e1ac060f612adcb454f6299f70583058cb508370"
     extract: False
 
-buildTools: [target-toolchain, curl]
-buildVars: [AUTOCONF_HOST, PKG_VERSION]
+buildTools: [target-toolchain, curl, host-toolchain]
+buildVars: [AUTOCONF_BUILD, AUTOCONF_HOST, PKG_VERSION]
 buildScript: |
+
+    EXTRA=( )
+    if cpackageCrossCompiling ; then
+        export ac_cv_fread_reads_directories=yes
+        export ac_cv_snprintf_returns_bogus=yes
+        EXTRA+=( "--build=${AUTOCONF_BUILD}" )
+    fi
+
     mkdir -p install
     tar -xf $1/git-${PKG_VERSION}.tar.xz
 
@@ -32,7 +40,9 @@ buildScript: |
                 --prefix=/usr \
                 --with-curl=${BOB_DEP_PATHS[net::curl-dev]}/usr \
                 --without-python \
-                --without-expat
+                --without-expat \
+                --without-iconv \
+                ${EXTRA:+"${EXTRA[@]}"}
     makeParallel LIB_4_CRYPTO="$(pkg-config --libs libssl libcrypto)"
     make DESTDIR=${PWD}/../install install
     popd


recipes/devel/git.yaml
```diff
@@ -21,18 +21,22 @@ checkoutSCM:
...
-    CPPFLAGS, CFLAGS, LDFLAGS]
+buildTools: [target-toolchain, curl, host-toolchain]
+buildVars: [PKG_VERSION, AUTOCONF_HOST, AUTOCONF_BUILD,
+    CC, CPPFLAGS, CFLAGS, LDFLAGS]
 buildScript: |
     export CFLAGS="-I${BOB_DEP_PATHS[libs::zlib-dev]}/usr/include -I${BOB_DEP_PATHS[libs::openssl-dev]}/usr/include"
     export LDFLAGS="-L${BOB_DEP_PATHS[libs::zlib-dev]}/usr/lib -L${BOB_DEP_PATHS[libs::openssl-dev]}/usr/lib -Wl,-rpath-link=${BOB_DEP_PATHS[libs::zlib-dev]}/usr/lib -Wl,-rpath-link=${BOB_DEP_PATHS[libs::openssl-dev]}/usr/lib"
 
+    export ac_cv_fread_reads_directories=yes
+    export ac_cv_snprintf_returns_bogus=no
```
Ferruck, line 32: As I've already mentioned in #142 this depends on the target libc. I don't think it's the proper solution to just hard-code this here. Once someone starts using something other than `glibc` with the basement this could easily lead to hard-to-debug problems.

Additionally, your changes hard-code these values even for native builds for no good reasons.

Ferruck (changes requested):

recipes/devel/git.yaml
mahaase: hm the second point we can handle easily, so we just do the `ac_` stuff for cross-compilation. the dependencies to glibc, mh, no idea: maybe we start with that and if we are on the point for that problem, we have to adjust this?!

Ferruck: > [...] maybe we start with that and if we are on the point for that problem, we have to adjust this?!

The problem I see with this approach is that users of the basement might use the `devel::git` recipe in a recipe stack with their own libc recipe. How are they supposed to know that they have to change the `devel::git` recipe in order for their setup to work properly? I mean, the compilation would still succeed and all...

Ferruck: In other words: I think we'd have to ensure that compilation of `devel::git` would fail if you try to compile it with a target libc other than `glibc`...

mahaase:
updated. 
i did the variables like buildroot:
https://github.com/buildroot/buildroot/blob/master/package/git/git.mk

recipes/devel/git.yaml
```diff
@@ -32,7 +40,9 @@ buildScript: |
                 --prefix=/usr \
                 --with-curl=${BOB_DEP_PATHS[net::curl-dev]}/usr \
                 --without-python \
-                --without-expat
+                --without-expat \
+                --without-iconv \
+                ${EXTRA:+${EXTRA[@]}}
```
jkloetzke, line 45: Without double quotes the entries in the array are subject to word splitting. I think this should be prevented:

```suggestion
                ${EXTRA:+"${EXTRA[@]}"}
```

recipes/devel/git.yaml
jkloetzke: I would be willing to see if this really ever happens. The current state of the PR only changes the behaviour for cross compiling. This should at least not make things worse. There are other places where we set such variables to hard coded values to get the package cross compiling. So it's not unheard of and so far it worked well enough.

jkloetzke (commented):
Just some minor thing...

jkloetzke:
Thanks!

PR #148, opened 2022-06-03
classes/cmake: set CMAKE_INSTALL_LIBDIR to lib

### LINKED ISSUE #141: set CMAKE_INSTALL_LIBDIR to lib statically
https://github.com/BobBuildTool/basement/blob/cbd2542abc66ba3df1636fb560681be1ab1623d2/classes/cmake.yaml#L142-L151

because of:

> CMAKE_INSTALL_LIBDIR
> object code libraries (lib or lib64 or lib/multiarch-tuple on Debian)
> -- https://cmake.org/cmake/help/v3.17/module/GNUInstallDirs.html

we have to set `CMAKE_INSTALL_LIBDIR` to `lib` to match our install.yaml functionality

https://github.com/BobBuildTool/basement/blob/cbd2542abc66ba3df1636fb560681be1ab1623d2/classes/install.yaml#L118-L126

mahaase:
.FIXES #141

### DIFF
--- a/classes/cmake.yaml
+++ b/classes/cmake.yaml
@@ -146,6 +146,7 @@ buildSetup: |
             -DCMAKE_BUILD_TYPE=Bob \
             -DCMAKE_INSTALL_PREFIX="$CMAKE_INSTALL_PREFIX" \
             -DCMAKE_INSTALL_SYSCONFDIR="$CMAKE_INSTALL_SYSCONFDIR" \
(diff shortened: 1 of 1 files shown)


jkloetzke:
Thanks. I've updated the commit message. Please always provide the rationale for the change in the commit message. Otherwise it's unclear in the future *why* this was changed.

PR #150, opened 2022-06-03
libs/compat/glibc: fixed the patchfile, added missing spaces

### LINKED ISSUE #149: glibc patch file not working
https://github.com/BobBuildTool/basement/blob/cbd2542abc66ba3df1636fb560681be1ab1623d2/recipes/libs/glibc/0001-glibc-po-files-out-of-tree.diff#L1-L21

i have no idea why, but this patch file doesn't works for me.

```
+ patch -p 1 -F 0 -f
patching file po/Makefile
Hunk #1 FAILED at 50.
Hunk #2 FAILED at 74.
2 out of 2 hunks FAILED -- saving rejects to file po/Makefile.rej
++ bob_handle_error 1
```

mahaase:
.FIXES #149


Ferruck:
In https://github.com/BobBuildTool/basement/issues/149#issuecomment-1145803006 you say that

> the **non-compat-one** fails currently!

now you're changing the compat one. I'm a bit confused...

daxcore:
argh. the compat glibc patch is failing for me. sorry for confusion! die PR solved the issue here.

jkloetzke:
I'm wondering how it built with the damaged patch in the first place... Anyway, thanks for the fix!

PR #153, opened 2022-06-14
Bugfix/msys2gitstuff

mahaase:
the git-stuff took us some days of searching for mismatching artifacts. the problem was, that a developer used the native git binary with autocrlf enabled. so the patches hat crlf endings, and the patched files differed to the other ones.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,2 @@
+.git*   eol=lf export-ignore
+*       eol=lf
--- a/tests/cross-platform/.gitignore
+++ b/tests/cross-platform/.gitignore
@@ -1,2 +1,4 @@
-/dev
+/dev*
 /work
+/graph
+/projects
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,2 @@
-/dev
-/work
 .bob-*
-user.yaml
 __pycache__/
--- a/tests/linux/.gitignore
+++ b/tests/linux/.gitignore
@@ -1,2 +1,4 @@
-/dev
+/dev*
 /work
+/graph
+/projects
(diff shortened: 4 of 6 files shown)


.gitattributes
```diff
@@ -0,0 +1,19 @@
+.git*           eol=lf export-ignore
+
+*.patch         eol=lf
```
jkloetzke, line 3: Wouldn't `*   eol=lf` be enough instead of listing each file type? I mean, there are no binary files anyway in this repository...

tests/cross-platform/.gitignore
```diff
@@ -1,2 +1,4 @@
-/dev
-/work
\ No newline at end of file
+/dev*
```
jkloetzke, line 1: Why the trailing `*`?

jkloetzke (commented):
Nice catch.

tests/cross-platform/.gitignore
mahaase: this is a best-practice for MSYS2/Windows, because deleting will take so much time, so we just rename it to dev_ before deleting it at background.
it is also usefull to backup the current dev workspace to get a clean one.

this is just a nice to have, not a strict requirement.

mahaase: this is a best-practice for MSYS2/Windows, because deleting will take so much time, so we just rename it to dev_ before deleting it at background.
it is also usefull to backup the current dev workspace to get a clean one.

this is just a nice to have, not a strict requirement.

jkloetzke: I see.

PR #154, opened 2022-06-16
generic sandbox: build a sandbox by available toolchain

### LINKED ISSUE #42: building in a qemu vm sandbox
feature request:

it would be nice, if it is possible to select e.g. aarch64 ARCH for sandbox to build "native" aarch64 on x86_64 system by using qemu-user-static.

currently we built a sandbox with that binary. the performance with qemu5 is impressive! 
so we build aarch64 artifacts (on x86_64 jenkins) for aarch64 targets (on which we also use bob natively! nvidia agx)

mahaase:
Update: Rebased onto current master (eb62005) and split into reviewable commits.

## What this does

Allows building a sandbox for an architecture that differs from the build
machine. The sandbox is then entered through qemu binfmt. Tested on
Ubuntu >= 18 LTS including WSL2.

Host setup to enter the sandbox:

- install `qemu-user-static` (`binfmt-support`)
- `qemu-user-static` may have to be installed manually from the ubuntu packages, it has no dependencies; **the higher the version the better the performance. it matters**!
- may need `update-binfmts --unimport && update-binfmts --import` plus a restart of the binfmt-support service

A foreign sandbox is described by three variables, see the
`cmake::greeter-cross-aarch64` example in `tests/cross-platform`:

```yaml
environment:
    SANDBOX_AUTOCONF_BUILD: "aarch64-bob-linux-gnu"   # triplet of the sandbox
    SANDBOX_ARCH: "arm64"                             # kernel architecture
    SANDBOX_AUTOCONF_TARGET: "aarch64-bob_compat-linux-gnu"
```

Note that building such a project with `--no-sandbox` will fail if the sandbox
architecture is not the host architecture.

## Enter The Sandbox Manually

```
host$ bob-namespace-sandbox -S work/devel/sandbox/dist/<id>/workspace -- /bin/bash

bash-5.3$ uname -a
Linux FDT-NB-0171 6.18.33.1-microsoft-standard-WSL2 #1 SMP PREEMPT_DYNAMIC Fri Jun  5 01:12:21 UTC 2026 aarch64 GNU/Linux
bash-5.3$ exit
exit

host$ rm -r work/devel/sandbox/dist/<id>/workspace/proc/
host$ rm -r work/devel/sandbox/dist/<id>/workspace/dev/
```

## Commits

| Commit | |
|---|---|
| `generic sandbox: build a sandbox by available toolchain` | the feature itself |
| `multiarch: add arm64 baseline for $(toolchain-feature)` | `$(toolchain-feature)` bailed out with *neither `$CROSS_TOOLCHAIN_ARCH` nor `$GCC_TARGET_ARCH` set* for `ARCH=arm64`, so an arm64 sandbox could not even be parsed. `x86_64` and `arm` already have a baseline, `arm64` now defaults to `armv8-a`. |
| `generic sandbox: drop redundant SANDBOX_AUTOCONF_HOST` | see below |
| `tests: drop SANDBOX_AUTOCONF_HOST from cross-platform example` | test side of the above |
| `cross-toolchain: use a distinct vendor for the aarch64 target` | independent fix, kept separate on purpose |

### Dropping SANDBOX_AUTOCONF_HOST

The first version introduced both `SANDBOX_AUTOCONF_HOST` and
`SANDBOX_AUTOCONF_BUILD`. They describe the same thing — the triplet of the
sandbox — and they must not diverge: the wrappers created by
`basement::bits::sandbox-toolchain` are named after `SANDBOX_AUTOCONF_BUILD`
and exec `/toolchain/bin/<AUTOCONF_TARGET>-*`. One knob is enough.

It also no longer sets `AUTOCONF_HOST` inside the sandbox:

- builds inside the sandbox are native with respect to the sandbox.
  `AUTOCONF_HOST` is the "we are cross compiling" marker that is exclusively
  provided by the target toolchain. Setting it made `classes/autotools.yaml`
  pass `--host=...`, so configure switched to cross compilation mode and
  skipped the run time checks of otherwise native builds.
- the `provideSandbox` environment took effect for exactly one package, the
  sandbox toolchain itself (everything after it picks up `AUTOCONF_BUILD` from
  its `provideVars`). But that is the one package that wants the *host*
  triplet: the wrapper fallback and the fingerprint script use
  `AUTOCONF_BUILD` to locate the host compiler when built without a sandbox.

As a result `classes/basement/bits/sandbox.yaml` is untouched by this PR now.

### The aarch64 target triplet

`$(host-autoconf)` returns `aarch64-linux-gnu` on an arm64 build machine, which
was exactly the `AUTOCONF_TARGET` of `devel::cross-toolchain-aarch64-linux-gnu`.
Wherever `AUTOCONF_BUILD` is the plain host triplet — any project using the
basement toolchains without `basement::rootrecipe` and its sandbox-toolchain —
`AUTOCONF_HOST == AUTOCONF_BUILD`, and the cross build silently degrades to a
native one (static instead of shared libs, no `--host=`, no `toolchain.cmake`).
The `x86_64-linux-gnu` toolchain already inserts a `cross` vendor for that
reason; this pulls aarch64 in line. `riscv64-linux-gnu` has the same problem
and is left alone for now.

**This is unrelated to the sandbox feature, so it is a separate commit at the end
— drop it if you would rather see it in its own PR. The `-ada` variant is left
untouched.**

## Verification

The default x86_64 sandbox is unaffected: the resolved
`buildVars`/`packageVars`/`buildTools` of the complete `cmake::greeter-host`
and `cmake::greeter-cross` trees of `tests/cross-platform` are byte-identical
to master (`bob show --sandbox --format flat`, diffed against a worktree of
eb62005). No rebuild for existing users.

For the aarch64 sandbox I only verified the package graph, **not an actual
build** — I have no arm64 binfmt setup here:

- sandbox content is cross built in the x86_64 bootstrap sandbox:
  `AUTOCONF_BUILD=x86_64-wendy-linux-gnu`, `AUTOCONF_HOST=aarch64-bob-linux-gnu`
- packages inside the finished sandbox: `AUTOCONF_BUILD=aarch64-bob-linux-gnu`,
  `AUTOCONF_HOST=aarch64-bob_compat-linux-gnu` — the same pattern as x86_64
- both sandbox toolchain packages keep the real host triplet
  (`x86_64-linux-gnu`)

A run of `cmake::greeter-cross-aarch64` on a qemu-binfmt enabled machine would
be good before merging.

Fixes #42

### DIFF
--- a/recipes/devel/cross-toolchain.yaml
+++ b/recipes/devel/cross-toolchain.yaml
@@ -88,7 +88,7 @@ multiPackage:
     # ARMv8-A AArch64 Linux toolchain
     aarch64-linux-gnu:
         environment:
-            AUTOCONF_TARGET: "aarch64-linux-gnu"
+            AUTOCONF_TARGET: "aarch64-cross-linux-gnu"
             ARCH: "arm64"
             GCC_TARGET_ARCH: "armv8-a"
             GCC_LIBC: "glibc"
--- a/classes/sandbox-toolchain.yaml
+++ b/classes/sandbox-toolchain.yaml
@@ -62,3 +62,4 @@ provideVars:
     # architecture is the same.
     AUTOCONF_BUILD: "${SANDBOX_AUTOCONF_BUILD}"
     AUTOCONF_HOST:  "${SANDBOX_AUTOCONF_HOST}"
+    ARCH: "${SANDBOX_ARCH:-$ARCH}"
--- a/classes/sandbox.yaml
+++ b/classes/sandbox.yaml
@@ -70,4 +70,7 @@ provideSandbox:
     mount:
         - "/etc/resolv.conf"
         - "/dev/pts"
+    environment:
+        AUTOCONF_HOST: "${SANDBOX_AUTOCONF_HOST:-${AUTOCONF_HOST}}"
+        AUTOCONF_BUILD: "${SANDBOX_AUTOCONF_BUILD:-${AUTOCONF_BUILD}}"
 
--- a/recipes/devel/bootstrap-sandbox.yaml
+++ b/recipes/devel/bootstrap-sandbox.yaml
@@ -41,7 +41,7 @@ depends:
     -
         name: devel::cross-toolchain
         environment:
-            AUTOCONF_TARGET: "x86_64-wendy-linux-gnu"
+            AUTOCONF_TARGET: "$(gen-autoconf,wendy)"
             ARCH: x86_64
             GCC_LIBC: glibc
             BASEMENT_OPTIMIZE: "s"
--- a/recipes/devel/host-compat-toolchain.yaml
+++ b/recipes/devel/host-compat-toolchain.yaml
@@ -6,7 +6,7 @@
 shared: True
 
 environment:
-    AUTOCONF_TARGET: "$(gen-autoconf,bob_compat)"
+    AUTOCONF_TARGET: "${SANDBOX_AUTOCONF_TARGET:-$(gen-autoconf,bob_compat)}"
 
 depends:
     # The following tools are needed by the cross-toolchain build process.
--- a/recipes/devel/sandbox-toolchain.yaml
+++ b/recipes/devel/sandbox-toolchain.yaml
@@ -1,5 +1,5 @@
 inherit: [sandbox-toolchain]
 
 privateEnvironment:
-    SANDBOX_AUTOCONF_BUILD: "$(gen-autoconf,bob)"
-    SANDBOX_AUTOCONF_HOST:  "$(gen-autoconf,bob)"
+    SANDBOX_AUTOCONF_BUILD: "${SANDBOX_AUTOCONF_BUILD:-$(gen-autoconf,bob)}"
+    SANDBOX_AUTOCONF_HOST:  "${SANDBOX_AUTOCONF_HOST:-$(gen-autoconf,bob)}"
--- a/recipes/devel/sandbox.yaml
+++ b/recipes/devel/sandbox.yaml
@@ -47,8 +47,8 @@ depends:
     -
         name: devel::cross-toolchain
         environment:
-            AUTOCONF_TARGET: "x86_64-bob-linux-gnu"
-            ARCH: x86_64
+            AUTOCONF_TARGET: "${SANDBOX_AUTOCONF_HOST:-$(gen-autoconf,bob)}"
+            ARCH: "${SANDBOX_ARCH:-$(host-arch)}"
             GCC_LIBC: glibc
             BASEMENT_OPTIMIZE: "s"
             BASEMENT_DEBUG: "0"
@@ -90,4 +90,3 @@ depends:
       depends:
         - devel::binutils
         - devel::gcc-native
-
--- a/tests/cross-platform/recipes/cmake/greeter.yaml
+++ b/tests/cross-platform/recipes/cmake/greeter.yaml
@@ -22,7 +22,7 @@ multiPackage:
             - cmake::libgreet-dev
             - cmake::libgreet-tgt
     cross:
-        depends:
+        depends: &cross
             - if: !expr |
                 "${BOB_HOST_PLATFORM}" == "msys" || "${BOB_HOST_PLATFORM}" == "win32"
               name: devel::win::vs2019-toolchain-vc142-amd64
@@ -34,3 +34,11 @@ multiPackage:
               use: [tools]
               forward: True
             - depends: *deps
+    cross-aarch64:
+        environment:
+            # this is just an example how to use this sandbox inside a recipe
+            SANDBOX_AUTOCONF_HOST: "aarch64-bob-linux-gnu"
+            SANDBOX_AUTOCONF_BUILD: "aarch64-bob-linux-gnu"
+            SANDBOX_AUTOCONF_TARGET: "aarch64-bob_compat-linux-gnu"
+            SANDBOX_ARCH: "arm64"
+        depends: *cross


mahaase:
hey. is there something wrong with this changes? i liked the aarch64 sandbox stuff... 🤔

jkloetzke:
There is nothing wrong with the change AFAICT. It just needs to be rebased. Then I can look again into the details...

recipes/devel/cross-toolchain.yaml
```diff
@@ -216,10 +216,10 @@ multiPackage:
                     GCC_LIBC: "uclibc-l4re"
                     GCC_TARGET_ARCH: "armv8-a"
 
-            # ARMv8-A AArch64 Linux toolchain
+            # ARMv8-A AArch64 "cross" compiling toolchain. Ships its own glibc.
             aarch64-linux-gnu:
                 environment:
-                    AUTOCONF_TARGET: "aarch64-linux-gnu"
+                    AUTOCONF_TARGET: "aarch64-cross-linux-gnu"
```
mahaase, line 222: @jkloetzke just updated this toolchain, left all other untouched! maybe some other ones like `-ada` needs the fix too?!

mahaase:
finished. test build finally run successfully!

mahaase:
works for me pretty well with the last adaptions.

jkloetzke:
Could you fold cc009a95565f6d4b72f9db5dd0a7b98f331bf571 and 6bae762376031c68b4fe05a7ee6dd7f72cb511b2 into the first commit? They are cleanups and I don't think they add to the understanding of the change.

I'm still torn about the last commit. Doesn't this mean that we should give all the cross compiling toolchains a separate vendor name?

mahaase:
> Could you fold [cc009a9](https://github.com/BobBuildTool/basement/commit/cc009a95565f6d4b72f9db5dd0a7b98f331bf571) and [6bae762](https://github.com/BobBuildTool/basement/commit/6bae762376031c68b4fe05a7ee6dd7f72cb511b2) into the first commit? They are cleanups and I don't think they add to the understanding of the change.

done.

> I'm still torn about the last commit. Doesn't this mean that we should give all the cross compiling toolchains a separate vendor name?

🤯... i had to ask 'someone' else ...

```
Short answer: no — only the ones whose triplet can actually alias the build machine's. 
But the commit as it stands is inconsistent, and there's one case it should have caught.

Why not all of them. The collision is a plain string compare in 
classes/basement/bits/libs.yaml:11 — AUTOCONF_BUILD != AUTOCONF_HOST. AUTOCONF_BUILD is 
$(host-autoconf) (default.yaml:20), and on Linux that function only ever produces <uname -m>-linux-gnu 
(plugins/multiarch.py:28-32, normalizing only AMD64→x86_64 and i*→i386). So a target triplet can only 
collide if it has exactly that four-field-less <machine>-linux-gnu shape. Anything carrying a distinguishing 
vendor, OS or ABI field is structurally safe:

┌───────────────────────────────────────┬─────────────────────────┬──────────────────────────────┐
│               toolchain               │     AUTOCONF_TARGET     │          collides?           │
├───────────────────────────────────────┼─────────────────────────┼──────────────────────────────┤
│ x86_64-linux-gnu, -ada                │ x86_64-cross-linux-gnu  │ fixed already                │
├───────────────────────────────────────┼─────────────────────────┼──────────────────────────────┤
│ aarch64-linux-gnu                     │ aarch64-cross-linux-gnu │ fixed by this commit         │
├───────────────────────────────────────┼─────────────────────────┼──────────────────────────────┤
│ aarch64-linux-gnu-ada                 │ aarch64-linux-gnu       │ yes — missed                 │
├───────────────────────────────────────┼─────────────────────────┼──────────────────────────────┤
│ riscv64-linux-gnu                     │ riscv64-linux-gnu       │ yes — deferred               │
├───────────────────────────────────────┼─────────────────────────┼──────────────────────────────┤
│ arm-linux-gnueabihf                   │ arm-linux-gnueabihf     │ no — host would be           │
│                                       │                         │ armv7l-linux-gnu             │
├───────────────────────────────────────┼─────────────────────────┼──────────────────────────────┤
│ *-none-eabi, *-none-elf,              │ —                       │ no                           │
│                                       │                         │ armv7l-linux-gnu             │
├───────────────────────────────────────┼─────────────────────────┼──────────────────────────────┤
│ *-none-eabi, *-none-elf,              │ —                       │ no                           │
│ *-unknown-elf, *-l4re                 │                         │                              │
├───────────────────────────────────────┼─────────────────────────┼──────────────────────────────┤
│ x86_64-w64-mingw32                    │ x86_64-w64-mingw32      │ only when building on        │
│                                       │                         │ MINGW64_NT                   │
└───────────────────────────────────────┴─────────────────────────┴──────────────────────────────┘

The thing worth acting on: aarch64-linux-gnu-ada (recipes/devel/cross-toolchain.yaml:228-233) still has 
the bare aarch64-linux-gnu and breaks in exactly the way the commit message describes. The x86_64 pair 
got the vendor on both the plain and the -ada variant; the aarch64 pair only got it on one. The commit 
message calls out riscv64 as a conscious deferral but says nothing about the ada variant, so this reads as 
an oversight rather than a decision.

What I'd do: extend the commit to cover aarch64-linux-gnu-ada and riscv64-linux-gnu. That gives you a 
statable invariant — a Linux/glibc target triplet must carry a vendor field, because otherwise it aliases the 
build machine's own triplet — instead of a set of individually-patched cases. Leave arm-linux-gnueabihf 
and the bare-metal/l4re/mingw ones alone: they can't collide, and changing them churns 
CROSS_COMPILE and the sysroot path for no benefit. Blanket-renaming everything would buy cosmetic 
uniformity at the cost of user-visible path changes.

Two things that would stop this recurring, if you want them:

- $(gen-autoconf,cross,<triplet>) already exists (plugins/multiarch.py:47) and does precisely this vendor 
  insertion. Using it instead of hand-spelled aarch64-cross-linux-gnu strings makes the intent explicit at 
  each site.
- A parse-time guard in the toolchain recipe that rejects AUTOCONF_TARGET == $(host-autoconf) would 
  catch the next ppc64le-linux-gnu someone adds. Caveat: it would also fire for mingw-on-mingw, so it'd 
  need a carve-out.
```

fixed the last commit as described. So what do you think now?

PR #156, opened 2022-06-21
devel/fakeroot: update to v1.29

mahaase:
v1.28 is deleted on the file server

.FIXES#155


jkloetzke:
Thanks.

BTW, I've slightly updated the commit message. Please follow the standard syntax with "Fixes #xxx."

mahaase:
oh! the space between fixes and id was missing :/ i guess `.FIXES #xxx` and `Fixes #xxx` are with the same functionality.

PR #160, opened 2022-12-04
Update core development packages

jkloetzke:


Ferruck:
Looks good to me. I'll build our current projects with it now and come back to you.

Ferruck:
Alas, our version of the i.MX ATF doesn't build with the updated toolchain and I currently have no means to update it. Thus, I cannot verify that everything else builds (nor that it works). But that's not your problem. Who knows when (or if!) we're going to update the basement.

One observation, though: When building multiple times, `devel::binutils` is sometimes rebuilt (and thus the GCC, as well). I'm unsure if this is caused by our environment or not, though.

Ferruck:
I'll test this again tomorrow using the `CFLAGS=+" --param=min-pagesize=0"` mentioned in the bug report. Sounds like the correct solution in *our* case. I'm not sure how to handle the existence of such a bug in general, though, especially since Bob is somewhat aimed at low-level stuff...

Ferruck:
> The toolchain is targeting Linux user space.

Yes, you're right. I didn't think of the `-linux-` infix in the toolchain's name. It would be different IMHO if the basement provided an AVR toolchain by default, but since this is not the case it's no show stopper (at least for me).

> Another option would be to patch gcc to default to `min-pagesize=0`.

I don't think that's a desirable solution.

I'm building our stuff right now, with the additional `CFLAG` included.

Ferruck:
I just stepped into a known issue caused by commit `7eae6a91e9b1670330c9f15730082c91c0b1d570` in the new glibc. There's probably a fix available but I'll need some time.

PR #161, opened 2022-12-18
libs/glibc: Backport header checks

Ferruck:
Without these some packages fail to compile because glibc redefines enums already defined by the kernel. The patches are upstream already and part of the next glibc release.

This fixes the issues I've mentioned in my [previous comment](https://github.com/BobBuildTool/basement/pull/160#issuecomment-1338986534).


jkloetzke:
I had to restart the CI build because it was interrupted by an unattended upgrade and apparently failed to resume...

PR #171, opened 2023-11-07
classes/cmake: add sysroot to cmake_find_root_path for all *gcc

### LINKED ISSUE #170: add sysroot to find_root_path for all *gcc
https://github.com/BobBuildTool/basement/blob/f27d80635d624c398982e21135b7bbfc626cbdd3/classes/cmake.yaml#L18

I guess we should check for:

```
if [[ "$CC" == *"gcc" ]] ; then
```

I have some recipes, where i had to add sysroot manually to find_root_path.

mahaase:
Fixes #170


jkloetzke:
Thanks!

PR #175, opened 2023-11-13
python/python3: prevent using system stuff while cross-compiling

mahaase:
without this workaround the system uses include directories of the host
like /usr/include/x86_64-linux-gnu and /usr/local/include

this results in build errors. root cause is the configure script
generated by autoconf:

```
  elif test "x$build_alias" != "x$host_alias"; then
    cross_compiling=yes
```

but in our case, using the host-compat-toolchain, host and build alias
is the same. cross-compiling will be disabled and system pahts used.

hope that fix is fine enough for general usage inside of bob.


jkloetzke:
I would call it a stop-gap solution. The real fix would be to have a better host-compat toolchain but this is much more involved. Anyway, let's use it for the moment. Thanks.

PR #176, opened 2023-11-13
utils/file: use https instead of ftp file download

mahaase:
at least in our network, the ftp url is not available.


jkloetzke:
Thanks!

PR #177, opened 2023-11-13
python/python3: python package includes library and headers

### LINKED ISSUE #168: python3 library
hey,

lovely to see the python3 support in bob basement https://github.com/BobBuildTool/basement/commit/ef077d2b4c1f0c26802faee0d6c842f29257ce73.

but how could we handle, if a package needs python3 library and header files?
this could not be realized with the current https://github.com/BobBuildTool/basement/blob/master/recipes/python/python3.yaml recipe.

do u have something in mind, how to support this?

in my first tests i realized, that i makes no sense to build a shared python3-dev package for target-toolchain.
in the end python3 tries to use some libraries while buildtime, which cant be resolved then. for host-toolchain, the python3 package will be implicitly static too.

so i guess we could extend the python3 recipe with:
```
privateEnvironment:
    BASEMENT_LIBS: "static"

[...]
        multiPackage:
            "":
                packageScript: |
                    autotoolsPackageTgt
                    ln -sf python3 usr/bin/python

            dev:
                packageScript: |
                    rsync -a "$1/install/" .
                    ln -sf python3 usr/bin/python

        provideTools:
            python3: usr/bin
```

and because static is forced, a tgt package makes no sense?!

**UPDATE**: that part with static lib, is wrong. what i tried to say is, that we may need a shared python3 library without dependencies to other shared libraries. so the dependencies have to be static, not python3 itself.

maybe something like this:
```
depends:
  - environment:
      BASEMENT_LIBS: "static"
    depends:
      - libs....
```

mahaase:
add python development support.

Fixes #168

### DIFF
--- a/recipes/python/python3.yaml
+++ b/recipes/python/python3.yaml
@@ -90,12 +90,15 @@ multiPackage:
             rsync -a "${BOB_DEP_PATHS[python::python3-wheel]}/" install/
 
         packageScript: |
-            autotoolsPackageTgt
+            autotoolsPackageBin
+            autotoolsPackageDev
             ln -sf python3 usr/bin/python
 
         provideDeps: [ "*-tgt" ]
         provideTools:
-            python3: usr/bin
+            python3:
+                path: usr/bin
+                netAccess: True
 
     minimal:
         depends:


recipes/python/python3.yaml
```diff
@@ -90,12 +90,15 @@ multiPackage:
...
+            autotoolsPackageBin
+            autotoolsPackageDev
             ln -sf python3 usr/bin/python
 
         provideDeps: [ "*-tgt" ]
         provideTools:
-            python3: usr/bin
+            python3:
+                path: usr/bin
+                netAccess: True
```
jkloetzke, line 101: I think this is too coarse. Every recipe that uses the "python3" tool will get network access. I think this should be restricted to where it's really needed. Calling pip and downloading from the internet is probably restricted to particular recipes. Would it be feasible to enable `buildNetAccess` there instead of centrally here?

recipes/python/python3.yaml
```diff
@@ -90,12 +90,15 @@ multiPackage:
             rsync -a "${BOB_DEP_PATHS[python::python3-wheel]}/" install/
 
         packageScript: |
-            autotoolsPackageTgt
+            autotoolsPackageBin
+            autotoolsPackageDev
```
jkloetzke, line 94: Hmm, now we always install the headers. Could there be separate -dev package instead? I have to admit I didn't think it through, though.

mahaase: now it looks like a common installation in a host system. installing the header will not have impact to the final image if we just install "*-tgt" in the end?

recipes/python/python3.yaml
jkloetzke: Exactly.

recipes/python/python3.yaml
jkloetzke: At least on Debian there are different packages for the python interpreter, the cpython headers and pip/setuptools. I think we should provide this distinction too. But I'm not sure what this means in our context and how it fits.

PR #181, opened 2023-12-11
Update python::python3-setuptools

tu-maurice:
To get a mildly recent setuptools-scm installed there needs to be a newer version of setuptools. So this updates setuptools to a version where the current build steps still work.


jkloetzke:
Thanks

PR #182, opened 2023-12-12
Add python::python3-dev with headers

tu-maurice:
Adds a separate package with /usr/include from the python3 build.


tu-maurice:
This seems to be adjacent to #168

PR #183, opened 2023-12-13
devel::git - Install using symlinks

tu-maurice:
git by default uses hardlinks to provide multiple executables with the same binary. This however doesn't translate well when copying around the installed files during the packaging step as the hardlinks are not reestablished, but the files are now duplicated, which results in a git installation with a size of more than 300MB.

For these cases git build system has a make variable called NO_INSTALL_HARDLINKS which can be passed along during make install to use symlinks instead of hardlinks.


tu-maurice:
Actually the copying command that duplicated the files for me is a local one. So maybe this PR is not necessary. I'll leave the decision with you.

jkloetzke:
I would actually like to leave the hardlinks. It has worked so far and is the usual install method. Thanks nonetheless.

PR #188, opened 2024-01-10
classes/python3-*: added the current pip install path to env var

### LINKED ISSUE #172: python3-usr: PYTHONPATH of current build dir is missing
https://github.com/BobBuildTool/basement/blob/9bd2604ecf81f0680597545967618f759828a158/classes/python3-usr.yaml#L7

the current build directory has just something like that:

`[libs::something-dev]=/invalid/exec/path/of/libs::something-dev`

so it is not possible to use pip installed stuff in the current buildScript.

UPDATE:
https://github.com/BobBuildTool/basement/blob/9bd2604ecf81f0680597545967618f759828a158/classes/python3-usr.yaml#L10

now i realized, that just existing paths are added. so we have to add them when pip install was called already!

in my blueprint/prototype i added all paths, if they are existing or not.

mahaase:
so it is possible to use python stuff, we installed in current build script.

.FIXES #172

### DIFF
--- a/classes/python3-pkg.yaml
+++ b/classes/python3-pkg.yaml
@@ -1,6 +1,14 @@
 inherit: [python3-usr]
 
 buildSetup: |
+    # add current pip install directory to PYTHONPATH for instant
+    # usage of packages installed by pip; inserting at front of
+    # list, so it is possible to use different version than system
+    for j in "" "lib-dynload" "site-packages" "dist-packages" ; do
+        PYTHONPATH="${BOB_CWD}/install${PYTHON3_DESTLIB}${j:+/$j}${PYTHONPATH:+:$PYTHONPATH}"
+    done
+    export PYTHONPATH
+
     # Make sure scripts do not hard code python3 path
     # $1: install path
     python3FixupScripts()


classes/python3-usr.yaml
```diff
@@ -12,6 +13,9 @@ buildSetup: &ref |
             fi
         done
     done
+    for j in "" "lib-dynload" "site-packages" "dist-packages" ; do
+        PYTHONPATH+="${PYTHONPATH:+:}${PYTHON3_INSTALL_DIR}${PYTHON3_DESTLIB}${j:+/$j}"
```
mahaase, line 17: that is not very clever. in case we would use a different version of a python package than the system one, this would not work. (e.g. setuptools)

better:
`PYTHONPATH="${PYTHON3_INSTALL_DIR}${PYTHON3_DESTLIB}${j:+/$j}${PYTHONPATH:+:$PYTHONPATH}"

BUT this feature is still missing!!

mahaase, line 17:
```suggestion
        PYTHONPATH="${PYTHON3_INSTALL_DIR}${PYTHON3_DESTLIB}${j:+/$j}${PYTHONPATH:+:$PYTHONPATH}"
```

jkloetzke: Which feature exactly is missing?

classes/python3-usr.yaml
```diff
@@ -12,6 +13,9 @@ buildSetup: &ref |
             fi
         done
     done
+    for j in "" "lib-dynload" "site-packages" "dist-packages" ; do
+        PYTHONPATH="${PYTHON3_INSTALL_DIR}${PYTHON3_DESTLIB}${j:+/$j}${PYTHONPATH:+:$PYTHONPATH}"
+    done
```
jkloetzke, line 18: Could you move this to the `python3-pkg` class instead?

mahaase: indeed, makes sense.

classes/python3-usr.yaml
mahaase: this one provided by this pull-request ;)

jkloetzke:
Thanks!

PR #190, opened 2024-01-17
devel/flex: fix regression introduced in flex 2.6.3

mahaase:
flex scanner: Use prefix when defining yywrap to avoid redefinition

### DIFF
--- a/recipes/devel/flex.yaml
+++ b/recipes/devel/flex.yaml
@@ -9,6 +9,10 @@ checkoutSCM:
     digestSHA256: 68b2742233e747c462f781462a2a1e299dc6207401dac8f0bbb316f48565c2aa
     stripComponents: 1
 
+checkoutDeterministic: True
+checkoutScript: |
+    patchApplySeries $<<flex/*>>
+
 buildTools: [bison]
 buildScript: |
     export M4=m4
--- a/recipes/devel/flex/0001-scanner-use-prefix-when-defining-yywrap-avoid-redefinition.patch
+++ b/recipes/devel/flex/0001-scanner-use-prefix-when-defining-yywrap-avoid-redefinition.patch
@@ -1,0 +1,27 @@
+From f5d87f1a26f4a5c3402497008ae10e9a1345d327 Mon Sep 17 00:00:00 2001
+From: Christos Zoulas <christos@zoulas.com>
+Date: Sun, 22 Jan 2017 18:20:44 +0100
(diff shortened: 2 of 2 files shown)


tu-maurice:
Actually... I think this might be the wrong patch? Let me try your PR with my local setup.

tu-maurice:
Yeah, no. This patch alone doesn't fix the issue. I've looked through all the commits in flex from v.2.6.3 to v.2.6.4 and imho the only other potentially relevant commit is from https://github.com/westes/flex/issues/142, which sadly doesn't apply cleanly to the v2.6.3 sources. 

flex v2.6.3 also seems to have caused a low of issues in other projects, so maybe we need a different approach: Any chance we could bump flex to v2.6.4 and also update doxygen to a newer version? It looks like doxygen v1.9.4 is the first release to require flex v2.6.4, which isn't too far off the current doxygen v1.9.1 in basement-gnu-linux.

jkloetzke:
I'm in favour of updating both projects if it helps with the problem. AFAICT there was no specific reason to use that version of doxygen. It just happened to be the one that got integrated back then.

mahaase:
I close this one and open a new PR.

tu-maurice:
Thanks.

mahaase:
@jkloetzke @tu-maurice 
I don't understand it. The ticket in doxygen issue tracker is closed, but I doesn't find a way to build doxygen successfully.

```
/bin/sh: line 1: 2626011 Segmentation fault      /bob/d90913cb3551492f51cfcc6a6ffb29887dfd0661/workspace/usr/bin/flex -o/bob/aea6ec06efc27d3ac75f5e18fa04319d2040cbc3/workspace/build/generated_src/code_intermediate.cpp /bob/aea6ec06efc27d3ac75f5e18fa04319d2040cbc3/workspace/build/generated_src/code.l
m4:stdin:838: ERROR: end of file in string
```

I used:

```
-- Found FLEX: /bob/d90913cb3551492f51cfcc6a6ffb29887dfd0661/workspace/usr/bin/flex (found version "2.6.4")
-- Found BISON: /bob/0fa8bcc2fd4b372d41f47d0b3eb482034676c695/workspace/usr/bin/bison (found version "3.8.2")
```

Before i also tried the upstream `bison`, but same issue.
Also the `-D_GNU_SOURCE` flag changed nothing, but I guess, I have a different issue, than this https://github.com/doxygen/doxygen/issues/8475.

PR #191, opened 2024-01-19
devel/flex: version bump to 2.6.4

mahaase:
fix included, that prevents segmentation fault while using flex

### DIFF
--- a/classes/basement/rootrecipe.yaml
+++ b/classes/basement/rootrecipe.yaml
@@ -35,6 +35,9 @@ depends:
         - name: devel::autotools
           use: [tools]
           forward: True
+        - name: devel::gettext
+          use: [tools]
+          forward: True
         - name: devel::bison
           use: [tools]
           forward: True
--- a/recipes/devel/bootstrap-sandbox.yaml
+++ b/recipes/devel/bootstrap-sandbox.yaml
@@ -59,7 +59,6 @@ depends:
     - devel::diffutils
     - devel::m4
     - devel::patch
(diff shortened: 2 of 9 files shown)


mahaase:
wait: WIP
this flex makes me crazy, one more bug. 
seems like can be fixed with that: https://github.com/westes/flex/issues/436#issuecomment-584495742

mahaase:
normally I would to patch the sources of flex, like e.g. archlinux does:
https://gitlab.archlinux.org/archlinux/packaging/packages/flex/-/blob/main/PKGBUILD?ref_type=heads 
but that is not possible, because autoconf isn't available while flex build.
any ideas? alternatively we could build with this flags:
```
# refer https://github.com/doxygen/doxygen/issues/8475
export CFLAGS+=" -D_GNU_SOURCE"
# refer https://github.com/westes/flex/issues/436
export ac_cv_func_malloc_0_nonnull=yes
export ac_cv_func_realloc_0_nonnull=yes
```
but no idea if that would break anything else?

jkloetzke:
The CI failure is caused by a missing "gettext" package on the host. Instead of adding it to the prerequisites, I would refactor the recipes so that gettext is used as a tool. Then we can even remove it from the sandbox. I'll prepare an update...

jkloetzke:
I've refactored the PR quite a bit to work around the build failure. Please have a look if it breaks anything on your side...

mahaase:
IMO we got it finally 👍

jkloetzke:
Great!

PR #192, opened 2024-01-19
devel/bison: version bump

mahaase:
not sure, if u want to merge it, it was created while searching the doxygen bug.
but no reason to update generally.


jkloetzke:
Sure, why not. Could you please update the commit message to include the version number too? The recommended format for all updates is "package: bump version to x.y.z"...

I'll merge it together with the host compat fix to reduce intermediate builds...

jkloetzke:
Thanks.

PR #199, opened 2024-03-14
bugfix python3 with openssl support

mahaase:
* bumped openssl to newest/last 1.1.1 version
* added `-DOPENSSL_THREADS` to `CFLAGS`

### DIFF
--- a/recipes/libs/openssl.yaml
+++ b/recipes/libs/openssl.yaml
@@ -1,7 +1,7 @@
-inherit: [cpackage, make, install]
+inherit: [cpackage, make, install, patch]
 
 metaEnvironment:
-    PKG_VERSION: "1.1.1i"
+    PKG_VERSION: "3.2.1"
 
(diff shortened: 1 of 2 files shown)


jkloetzke:
I'm afraid this is not the right solution. The `OPENSSL_THREADS` define should be provided by OpenSSL headers. Could you try an update to OpenSSL 3.2.1? That should (hopefully) require no changes to python recipe.

mahaase:
I tested 3.2.1 already. it build just fine, but however the OPENSSL_THREADS was not provided automatically. I will double check this!

jkloetzke:
Interesting. I checked the sources yesterday and could find it. Maybe we're building OpenSSL not correctly.

mahaase:
fixed it. added patch & bumped version. reverted python3 recipe changes.

jkloetzke:
Thanks!

PR #203, opened 2024-07-09
Genimage

sixtyfourktec:
Add genimage + dependency to the basement layer.


sixtyfourktec:
It was decided this fits more into basement-gnu-linux. So I created a pull request there.

PR #204, opened 2024-07-09
cpackage: add cpackageLibraryTypeCheck convenience function

sixtyfourktec:
The cpackageLibraryTypeCheck function can be used to get a correct list of input items based on the library type being build. It has two input parameters, the first being the item for the static library case and the second parameter describing an item for the shared library case. Use it like this:

BUILD_TARGETS=$(cpackageLibraryTypeCheck libstatic.a libshared.so)


jkloetzke:
Thanks

PR #207, opened 2024-07-12
Fix deprecated warnings when using meson

sixtyfourktec:
Fix some warnings from meson about features we use which are deprecated.


jkloetzke:
Thanks

PR #208, opened 2024-08-08
autoconf: Add L4Re OS

admlck:
Add the L4Re OS to the config.sub such that it is added to projects when they run autoconf.


jkloetzke:
Thanks!

PR #209, opened 2024-08-09
Add fakeroot library symlink

sixtyfourktec:
Fakeroot opens its library manually by looking for libfakeroot.so. However, the installed library name is libfakeroot-0.so. autotoolsPackageTgt ignores the installed symlink, because the SONAME is libfakeroot-0. Add the symlink back, so the fakeroot binary finds the library.


jkloetzke:
Thanks.

PR #210, opened 2024-08-19
initial Ada support

rhubert:
This adds support for building a cross-toolchain with ada-language enabled. 

As a gnat-compiler is needed to build gcc with gnat this needs to be setup before. There is a ada-layer at https://github.com/rhubert/layers-ada which uses `alr` to download the gnat-compiler from adacore. 
Using the gnat-toolchain from this layer will not change the target / host-toolchain setting. Instead it provides a `gnat-host-toolchain` and a `gnat-target-toolchain` (based on gcc-13.2). 

This is because we still want to use the `host-compat-toolchain` for all non-ada components - otherwise these tools won't be run-able on older systems.
OTOH we don't want to build + use the `host-compat-toolchain` with ada and use this for the ada-host-tools because they are using the latest ada-features + fixes.

To build a ada-enabled cross toolchain use:
```yaml
inherit: [basement::rootrecipe]

depends:
    - name: adacore::alr
      use: [tools]
      forward: True

    - name: adacore::gnat-x86_64-pc-linux-gnu
      use: [tools]
      forward: True

    - name: devel::cross-toolchain-aarch64-linux-gnu-ada    
      use: [tools]    
      forward: True    
      tools:    
        target-toolchain: gnat-host-toolchain    
        host-native-toolchain: gnat-host-toolchain    
```

### DIFF
--- a/recipes/devel/binutils.yaml
+++ b/recipes/devel/binutils.yaml
@@ -12,31 +12,53 @@ checkoutSCM:
 # Some parts are compiled for the host during compilation. Hence we need the
 # host toolchain too.
 buildTools: [host-toolchain, bison]
+multiPackage:
+    "":
+        buildVars: [AUTOCONF_HOST, AUTOCONF_TARGET, BINUTILS_PREFIX]
+        buildScript: |
+            export MAKEINFO=true
+            autotoolsBuild -o MAKEINFO=true -O MAKEINFO=true $1 \
+                ${BINUTILS_PREFIX:+--prefix=${BINUTILS_PREFIX}} \
+                --with-sysroot=${BINUTILS_PREFIX:-/usr}/sysroots/${AUTOCONF_TARGET:-$AUTOCONF_HOST} \
+                --enable-deterministic-archives \
+                --disable-nls \
+                --disable-werror \
+                --enable-plugins \
+                --enable-lto
 
-buildVars: [AUTOCONF_HOST, AUTOCONF_TARGET, BINUTILS_PREFIX]
-buildScript: |
-    export MAKEINFO=true
-    autotoolsBuild -o MAKEINFO=true -O MAKEINFO=true $1 \
-        ${BINUTILS_PREFIX:+--prefix=${BINUTILS_PREFIX}} \
-        --with-sysroot=${BINUTILS_PREFIX:-/usr}/sysroots/${AUTOCONF_TARGET:-$AUTOCONF_HOST} \
-        --enable-deterministic-archives \
-        --disable-nls \
-        --disable-werror \
-        --enable-plugins \
-        --enable-lto
+        multiPackage:
+            dev:
+                packageScript: |
+                    autotoolsPackageDev
+            "":
+                packageScript: |
+                    autotoolsPackageTgt
 
-packageScript: |
-    autotoolsPackageTgt
+                    # make sure that prefixed versions are available on native build
+                    if [[ $AUTOCONF_HOST = ${AUTOCONF_TARGET:-$AUTOCONF_HOST} ]] ; then
+                        pushd ./${BINUTILS_PREFIX:-/usr}/bin
+                        for i in *; do
+                            ln -s "$i" "${AUTOCONF_HOST}-$i"
+                        done
+                        popd
+                    fi
 
-    # make sure that prefixed versions are available on native build
-    if [[ $AUTOCONF_HOST = ${AUTOCONF_TARGET:-$AUTOCONF_HOST} ]] ; then
-        pushd ./${BINUTILS_PREFIX:-/usr}/bin
-        for i in *; do
-            ln -s "$i" "${AUTOCONF_HOST}-$i"
-        done
-        popd
-    fi
+                provideTools:
+                    binutils: usr/bin
 
-provideTools:
-    binutils: usr/bin
+    libiberty:
+        depends:
+            - devel::binutils-dev
+        buildScript: |
+            autotoolsBuild $1/libiberty \
+                --enable-install-libiberty
+            if [[ -e install/usr/lib64 ]]; then
+                mkdir -p install/usr/lib
+                mv install/usr/lib64/* install/usr/lib
+            fi
+        multiPackage:
+            dev:
+                packageScript: autotoolsPackageDev
+            tgt:
+                packageScript: autotoolsPackageTgt
 
--- a/recipes/devel/cross-toolchain.yaml
+++ b/recipes/devel/cross-toolchain.yaml
@@ -106,6 +106,15 @@ multiPackage:
             GCC_TARGET_ARCH: "armv8-a"
             GCC_LIBC: "glibc"
 
+    # ARMv8-A AArch64 Linux toolchain with ada
+    aarch64-linux-gnu-ada:
+        environment:
+            AUTOCONF_TARGET: "aarch64-linux-gnu"
+            ARCH: "arm64"
+            GCC_TARGET_ARCH: "armv8-a"
+            GCC_LIBC: "glibc"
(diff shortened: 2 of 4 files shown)


jkloetzke:
> @jkloetzke : Do you think the ada-layer could become a BobBuildTool/ada - layer?

Certainly, especially since there seems to be some interdependency between the layers.

recipes/devel/binutils.yaml
```diff
@@ -12,31 +12,50 @@ checkoutSCM:
...
+                        for i in *; do
+                            ln -s "$i" "${AUTOCONF_HOST}-$i"
+                        done
+                        popd
+                    fi
+
+                provideTools:
+                    binutils: usr/bin
+
+    libiberty-dev:
```
jkloetzke, line 49: Shouldn't there also be a `libiberty-tgt`?

recipes/devel/m4.yaml
```diff
@@ -15,3 +15,5 @@ buildScript: |
 packageScript: |
     autotoolsPackageTgt
 
+provideTools:
+    m4: "usr/bin"
```
jkloetzke, line 19: Hmm, so far `m4` is a basement build prerequisite. It has to be installed on the host and is available in the sandbox for that reason. Is this not sufficient?

If we provide it as tool we may want to do this consistently. That is, also drop it from the sandbox...

classes/kernel-dev.yaml
```diff
@@ -0,0 +1,211 @@
+# Development source linux kernel. When built, this recipe packages the \
```
jkloetzke, line 1: Basically this should be part of basement-gnu-linux. Even though we technically have a kernel in the basement repository it's only here to have the libc kernel headers. But on another note:

* why a class and not a multipackage on the kernel::linux recipe?
* recent kernels have the `modules_prepare` build target. Would that help or be usable in your case?
* how is this related to the Ada support? I'd rather keep unrelated things separated in different PRs...

@sixtyfourktec Any more thoughts on this?

jkloetzke (commented):
It looks like that most of the commits are unrelated to Ada, are they? I'd prefer to have a separate PR for those commits...

recipes/devel/binutils.yaml
```diff
@@ -12,31 +12,53 @@ checkoutSCM:
...
+                        done
+                        popd
+                    fi
+
+                provideTools:
+                    binutils: usr/bin
+
+    libiberty:
+        depends:
+            - devel::binutils-dev
```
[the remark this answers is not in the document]
rhubert: I need both, `binutils-dev` and `libiberty-dev` for ada-bfd:

https://github.com/rhubert/layers-ada/blob/master/recipes/ada/ada-bfd.yaml#L14 

binutils-dev:
```
dev/dist/devel/binutils-dev/x86_64-bob_compat_gnat-linux-gnu/1/workspace$ tree
.
└── usr
    ├── include
    │   ├── ansidecl.h
    │   ├── bfd.h
    │   ├── bfdlink.h
    │   ├── collectorAPI.h
    │   ├── ctf-api.h
    │   ├── ctf.h
    │   ├── diagnostics.h
    │   ├── dis-asm.h
    │   ├── libcollector.h
    │   ├── libfcollector.h
    │   ├── plugin-api.h
    │   ├── sframe-api.h
    │   ├── sframe.h
    │   └── symcat.h
    └── lib
        ├── libbfd.a
        ├── libctf.a
        ├── libctf-nobfd.a
        ├── libgprofng.a
        ├── libopcodes.a
        └── libsframe.a

4 directories, 20 files
```

libiberty-dev:
```
dev/dist/devel/binutils-libiberty-dev/x86_64-bob_compat_gnat-linux-gnu/1/workspace$ tree
.
└── usr
    ├── include
    │   └── libiberty
    │       ├── ansidecl.h
    │       ├── demangle.h
    │       ├── dyn-string.h
    │       ├── fibheap.h
    │       ├── floatformat.h
    │       ├── hashtab.h
    │       ├── libiberty.h
    │       ├── objalloc.h
    │       ├── partition.h
    │       ├── safe-ctype.h
    │       ├── sort.h
    │       ├── splay-tree.h
    │       └── timeval-utils.h
    └── lib
        └── libiberty.a

5 directories, 14 files
```

There are two reasons for having this separate:
- libiberty is also included in gcc. I'm unsure about the consequences of building binutils with `--enable-libiberty` unconditionally.
- when I started writing recipes for the actual OSS-Project I did this in a VM on a thermal throttled machine having the power of a pocked calculator.  Changing the `binutils` package results in a rebuild of nearly everything occupying the "machinchen" for > 1day.

classes/kernel-dev.yaml
rhubert: > Basically this should be part of basement-gnu-linux. Even though we technically have a kernel in the basement repository it's only here to have the libc kernel headers. But on another note:
> 
>     * why a class and not a multipackage on the kernel::linux recipe?
> 
Cause it can than be used in different kernel recipes. (vendor kernels, ...)

>     * recent kernels have the `modules_prepare` build target. Would that help or be usable in your case?
> 
AFAICR this didn't work. But I can't remember why. 

>     * how is this related to the Ada support? I'd rather keep unrelated things separated in different PRs...
> 
The name of the PR is outdated.

I can remove this class and put it in a different layer...

jkloetzke:
> > It looks like that most of the commits are unrelated to Ada, are they? I'd prefer to have a separate PR for those commits...
> 
> Hmm...well they are unrelated to ada but I need all of them to build my project. I could split them into separate PR's but this would increase my effort on every change requested...

I don't think we need to split this into more than two or three PRs. What I'd like to have is this PR to add the Ada support and at least another one adding the unrelated packages. To be precise, I think most of the new recipes should go to the basement-gnu-linux layer because they are out of scope of basement. Remember, the basement layer should bring the sandbox, the toolchains and the infrastructure to build with the standard build systems. Of course, this includes the required dependencies but everything else should go to a different layer.

So, additionally to the Ada support, adding `classes/python3-build.yaml` and its dependencies seems sensible. The other new python modules and utils should certainly go to basement-gnu-linux instead...

rhubert:
> So, additionally to the Ada support, adding `classes/python3-build.yaml` and its dependencies seems sensible. The other new python modules and utils should certainly go to basement-gnu-linux instead...

It's in basement because I don't see why it shouldn't work on non gnu-linux as well?

recipes/devel/binutils.yaml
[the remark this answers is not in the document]
jkloetzke: > * libiberty is also included in gcc. I'm unsure about the consequences of building binutils with `--enable-libiberty` unconditionally.

Argh, this libiberty seems to be a complete mess. This library is bundled together with other GNU packages and should apparently have never been used by others. :facepalm: 

> * when I started writing recipes for the actual OSS-Project I did this in a VM on a thermal throttled machine having the power of a pocked calculator.  Changing the `binutils` package results in a rebuild of nearly everything occupying the "machinchen" for > 1day.

I see. Well, I guess we can restructure the package if this makes sense in the future.

jkloetzke:
Well, in retrospect the name "basement-gnu-linux" was not really the brightest idea. :roll_eyes: Of course, these python modules might also work on other operating systems. But that is not the point. I really want to keep the basement layer as slim as possible. Everything above the standard build system support should be in a different layer.

My hope is that basement will be a stable foundation in the future which is only receiving mostly toolchain updates. All these other modules, utilities and libraries that will create a lot of churn which should be somewhere else. Also, as we expand CI tests for basement, having all those unrelated libraries here will make the life harder. Because any update to the core basement recipes usually trigger a rebuild of a substantial amount of packages, if not all.

jkloetzke:
Thanks!

PR #211, opened 2024-08-28
openssl version bump and cleanup

admlck:
This bumps openssl to 3.3.1 and removes a variable assignment that is not needed anymore.


jkloetzke:
Thanks!

PR #212, opened 2024-08-29
zlib: Upgrade to 1.3.1

admlck:


jkloetzke:
Thanks.

PS: the established commit message pattern would have been "libs/zlib: bump to version 1.3.1" but that is not a big deal...

PR #213, opened 2024-08-30
Enhancements and L4Re support

admlck:


jkloetzke:
Apparently, L4Re has been committed to the GNU config helpers. :tada: I've prepared #214 that should consistently update the respective files in all packages. If that works, we can drop the config.sub patches for good.

admlck:
Wohooo!!!

jkloetzke:
I've removed the `configure.ac` patch because this caused the bash compilation problems. If `configure` is older than `configure.ac`, the `Makefile` will trigger an autoconf run. This will modify the sources. Luckily, autoconf is not available in the CI environment...

PR #219, opened 2024-09-17
Rust

sixtyfourktec:
Add rust/cargo build support.

### DIFF
--- a/recipes/devel/rustc.yaml
+++ b/recipes/devel/rustc.yaml
@@ -1,0 +1,81 @@
+inherit: [install]
+
+metaEnvironment:
+    PKG_VERSION: "1.81.0"
+    PKG_LICENSE: "Apache-2.0 or MIT"
+
+multiPackage:
+    # rust compiler + std lib + tools like cargo
+    "":
+        environment:
+            RUST_ARCH: $(rustc-target,$(get-tool-env,"target-toolchain","AUTOCONF_HOST"))
+
+        checkoutSCM:
+            scm: url
+            url: https://static.rust-lang.org/dist/rust-${PKG_VERSION}-${RUST_ARCH}.tar.xz
+            digestSHA1: >-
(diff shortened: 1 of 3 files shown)


recipes/devel/rustc.yaml
```diff
@@ -0,0 +1,155 @@
...
+
+privateEnvironment:
+    RUST_MIRROR: "https://static.rust-lang.org"
+
+multiPackage:
+    # rust compiler + std lib + tools like cargo
+    "":
+        # provide the rustc package based on the selected target architecture
+        depends:
+            - "devel::rustc-$(get-tool-env,target-toolchain,ARCH)"
```
jkloetzke, line 15: In the long run, `$ARCH` is certainly not sufficient. But that doesn't mean we must already roll the generic solution either. So until other OSes are supported, we might want to name the variants accordingly, e.g. `devel::rustc-x86_64-unknown-linux-gnu` -> `"devel::rustc-$(get-tool-env,target-toolchain,ARCH)-unknown-linux-gnu"`. This would kill the arm variant but it is better than requiring to refactor other recipes in the long run...

sixtyfourktec: This would also kill the aarch64 variant (which I need). Should I try to use some ifelse magic or is it really worth that effort?

sixtyfourktec: Actually it would also kill the i386 variant ;) I would prefer to wait until we exactly know how this would look like for different OS's.

sixtyfourktec: Well, it removed some functionality when installing rust on the target. On the previous change you could install the rust compiler on the target + any extra std libs you want for doing cross compilation on the target. This is not possible anymore. Also are those if expr in the depends section possible in the current version of bob?

sixtyfourktec: Sounds ok to me.

sixtyfourktec:
I tested this and made some small fixes and added your proposal.

jkloetzke:
Thanks!

PR #222, opened 2024-09-25
Exclude m4 files when installing on target

sixtyfourktec:
Those are configure extensions and only required when building.


jkloetzke:
I guess we have to do it the other way around by removing the m4 files in packages that install them...

PR #223, opened 2024-09-26
sandbox: don't remove the code page converting libs

sixtyfourktec:
Actually some tools require code page converting functionality when used as host tools. One of those are the mtools which try to convert to some fat specific code pages and fail when the gconv libraries are not there.


jkloetzke:
Thanks.

PR #224, opened 2024-09-27
python: make useable in checkoutscript + bump setuptools to 75.1.0

rhubert:


classes/python3-usr.yaml
```diff
@@ -1,5 +1,5 @@
-buildTools: [python3]
-buildSetup: &ref |
+checkoutTools: [python3]
```
jkloetzke, line 1: The `checkoutSetup` doesn't hurt. But the `checkoutTools` will affect _all_ python packages and potentially create distinct checkouts where in fact this is undesired. Would that be an option to add the `checkoutTools: [python3]` only to affected recipes?

Also, what are "meson submodules"?

rhubert: Sure, I updated the PR 


~submodules~ -> subprojects: https://mesonbuild.com/Subprojects.html  some meson magic, especially in combination with `wrap` files https://mesonbuild.com/Wrap-dependency-system-manual.html

Found this in qemu: https://github.com/Xilinx/qemu/tree/master/subprojects

jkloetzke:
Thanks.

PR #226, opened 2024-10-01
libs::glibc: use GNU_MIRROR

rhubert:


jkloetzke:
Good catch. Thanks.

rhubert:
ftp.gnu.org was down this morning...

jkloetzke:
Holy cow! :astonished:

PR #227, opened 2024-10-04
core::util-linux: bump to v2.40.2

sixtyfourktec:
Add a new generic bin target and install only libs in *-tgt. Some apps depend on the libs and don't necessarily need the executables as well. Also there was a provideDeps missing.

### DIFF
--- a/recipes/core/util-linux.yaml
+++ b/recipes/core/util-linux.yaml
@@ -2,7 +2,7 @@ inherit: [autotools]
 
 metaEnvironment:
     PKG_VERSION_MAJOR: "2.40"
-    PKG_VERSION: "2.40.1"
+    PKG_VERSION: "2.40.2"
 
 depends:
     - libs::ncurses-dev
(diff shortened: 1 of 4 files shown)


recipes/core/util-linux.yaml
```diff
@@ -34,11 +34,19 @@ buildScript: |
...
     dev:
         packageScript: autotoolsPackageDev
+        provideDeps: [ "*-dev" ]
 
     tgt:
-        packageScript: autotoolsPackageTgt
+        packageScript: autotoolsPackageLib
         provideDeps: [ "*-tgt" ]
 
     tools:
```
sixtyfourktec, line 52: Actually we also could remove the extra tools target and move this into the generic bin target. What do you think?

jkloetzke: Yes, this sounds reasonable. :+1:

recipes/core/util-linux.yaml
```diff
@@ -2,7 +2,7 @@ inherit: [autotools]
 
 metaEnvironment:
     PKG_VERSION_MAJOR: "2.40"
-    PKG_VERSION: "2.40.1"
+    PKG_VERSION: "2.40.2"
```
jkloetzke, line 5: The version bump should go into a separate commit.

recipes/core/util-linux.yaml
sixtyfourktec: done

jkloetzke:
Thanks!

PR #228, opened 2024-10-08
Addendum to 32e9ecf

sixtyfourktec:
rsync doesn't consider /usr/share/pkgconfig if not also /usr/share is specified.


jkloetzke:
Thanks.

PR #233, opened 2024-11-21
add pre-commit

rhubert:
Add initial pre-commit configuration files. The pre-commit tool can be
    installed using pip:
       pip3 install pre-commit
    
    After that the hook can be activated using:
       pre-commit install
    
    or just run on the modified files
       pre-commit run
    
    ATM the following checks are active:
      - check-yaml
      - end-of-file-fixer
      - trailing-whitespace
      - yamllint
      - codespell

### DIFF
--- a/.yamllint
+++ b/.yamllint
@@ -1,0 +1,54 @@
+---
+
+yaml-files:
+    - '*.yaml'
+    - '.yamllint'
+
+rules:
+    anchors:
+        forbid-undeclared-aliases: true
+        forbid-duplicated-anchors: true
+        forbid-unused-anchors: true
+    braces:
+        forbid: non-empty
+    brackets:
+        forbid: false
+        min-spaces-inside: -1
+        max-spaces-inside: -1
+        min-spaces-inside-empty: -1
+        max-spaces-inside-empty: -1
+    colons:
+        max-spaces-before: 0
+        max-spaces-after: -1
+    commas: enable
+    comments:
+        require-starting-space: true
+        ignore-shebangs: true
+        min-spaces-from-content: 1
+    comments-indentation: enable
+    document-end: disable
+    document-start: disable
+    empty-lines:
+        max: 1
+        max-start: 0
+        max-end: 0
+    empty-values: disable
+    float-values: disable
+    hyphens: enable
+    indentation:
+        spaces: 4
+        indent-sequences: consistent
+        level: warning
+    key-duplicates:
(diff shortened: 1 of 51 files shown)


jkloetzke:
Thanks. I don't think the `pyproject.toml` file is necessary. I pushed an update that moves it to `.codespellrc`. I hope that's ok.

rhubert:
Thanks. :+1:

I just found out that `regex_exclude` setting of `yamlfmt` is somehow not what I want as it excludes the file if the regex matches. Removes all whitespaces (before comments, between `key:   value` pairs,...) making the yaml files look mostly ugly. :( Since the configuration options are very limited I think we won't use `yamlfmt`... Maybe `yamllint` to enable some checks + fix them by hand?

jkloetzke:
> Maybe `yamllint` to enable some checks + fix them by hand?

Sounds reasonable. I guess we can fix the findings incrementally. The pre-commit thingy seems to support only looking at changed files.

.yamllint
jkloetzke, line 1: As this file is checked too, the indentation should probably be correct upfront.

.yamllint
```diff
@@ -0,0 +1,58 @@
...
+    min-spaces-from-content: 1
+  comments-indentation: enable
+  document-end: disable
+  document-start: disable
+  empty-lines:
+    max: 1
+    max-start: 0
+    max-end: 0
+  empty-values: disable
+  float-values: enable
```
jkloetzke, line 40: I think we can disable float values. Bob does not use them. Or do you use some plugin that uses float values?

.yamllint
```diff
@@ -0,0 +1,58 @@
...
+  anchors:
+    forbid-undeclared-aliases: true
+    forbid-duplicated-anchors: true
+    forbid-unused-anchors: true
+  braces:
+    forbid: false
+    min-spaces-inside: -1
+    max-spaces-inside: -1
+    min-spaces-inside-empty: -1
+    max-spaces-inside-empty: -1
```
jkloetzke, line 17: I would suggest:

```yaml
  braces:
    forbid: non-empty
```

Braces are not used in recipes. The only exception is `recipes/devel/cross-toolchain.yaml:73` where it's required.

jkloetzke (commented):
Yeah, yamllint seems to be much more useful.

jkloetzke:
Thanks! I'll try to add the check to the CI build in the next days.

PR #234, opened 2024-11-22
plugins/msbuild: fix SyntaxWarning

rhubert:
Use a raw-string here to avoid

 plugins/msbuild.py:18: SyntaxWarning: invalid escape sequence '\*'
  '-find', 'MSBuild\**\Bin\MSBuild.exe',


jkloetzke:
Thanks

PR #235, opened 2024-11-22
Python wheel

rhubert:
We had two different versions of "wheel". Update the basement-version and remove it from basement-gnu-linux.


jkloetzke:
I would propose to rename `python::python3-wheel` to `python::wheel` like it was in basement-gnu-linux. The `python3-` prefix is kind of moot since we do not support Python2 any more.

jkloetzke:
Thanks

PR #242, opened 2024-12-17
libs::libc: add openmp-tgt target

sixtyfourktec:
Make it possible to install gcc's libgomp.so into the target. Some applications may use it.


jkloetzke:
Thanks

PR #247, opened 2025-02-02
Remove build prequisites

rhubert:
On some systems not all build prequisites are installed / can be installed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,6 @@ Docs](https://bob-build-tool.readthedocs.io/projects/basement/en/latest/index.ht
   * gcc >= 5.x
   * bash
   * POSIX awk (GNU awk version >= 3.1.5)
-  * GNU m4
   * GNU make >= 3.80
   * perl >= 5.6.1
   * GNU tar
--- a/classes/basement/rootrecipe.yaml
+++ b/classes/basement/rootrecipe.yaml
@@ -32,6 +32,9 @@ depends:
           - name: devel::make
             use: [tools]
             forward: True
+          - name: devel::m4
+            use: [tools]
+            forward: True
           - name: devel::autotools
             use: [tools]
             forward: True
--- a/classes/basement/rootrecipe.yaml
+++ b/classes/basement/rootrecipe.yaml
@@ -35,6 +35,9 @@ depends:
           - name: devel::m4
             use: [tools]
             forward: True
+          - name: devel::help2man
+            use: [tools]
+            forward: True
           - name: devel::autotools
             use: [tools]
             forward: True
--- a/classes/basement/rootrecipe.yaml
+++ b/classes/basement/rootrecipe.yaml
@@ -38,6 +38,9 @@ depends:
           - name: devel::help2man
             use: [tools]
             forward: True
+          - name: devel::texinfo
+            use: [tools]
+            forward: True
           - name: devel::autotools
             use: [tools]
             forward: True
--- a/recipes/devel/help2man.yaml
+++ b/recipes/devel/help2man.yaml
@@ -1,0 +1,19 @@
+inherit: [autotools]
+metaEnvironment:
+    PKG_VERSION: "1.49.3"
+
+checkoutSCM:
+    scm: url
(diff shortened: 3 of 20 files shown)


recipes/devel/bootstrap-fake-m4.yaml
jkloetzke, line 1: This is actually a `bootstrap-fake-autotools`, isn't it?

recipes/devel/help2man.yaml
```diff
@@ -0,0 +1,44 @@
...
+       static)
+          SHARED_STATIC=( "--disable-shared" "--enable-static" )
+          ;;
+       shared)
+          SHARED_STATIC=( "--enable-shared" "--disable-static" )
+          ;;
+    esac
+
+    if [[ $1/configure -nt .configure.stamp ]] ; then
+        $1/configure \
```
jkloetzke, line 26: Couldn't we use the autotools class instead of replicating it's behaviour?

recipes/devel/autotools.yaml
```diff
@@ -7,6 +7,7 @@ depends:
     - devel::autoconf-archive
     - devel::automake
     - devel::libtool
+    - devel::m4
```
jkloetzke, line 10: Why is this necessary? Previously m4 had to be available on the host or in the sandbox. Now it's available as tool. Why do we need another copy?

recipes/devel/autotools.yaml
```diff
@@ -8,6 +8,7 @@ depends:
     - devel::automake
     - devel::libtool
     - devel::m4
+    - devel::texinfo
```
jkloetzke, line 11: Likewise. It should rather be added as buildTool where necessary, shouldn't it?

classes/basement/rootrecipe.yaml
```diff
@@ -16,6 +16,9 @@ depends:
           - name: devel::bootstrap-fake-pkg-config
             use: [tools]
             forward: True
+          - name: devel::bootstrap-fake-autotools
```
jkloetzke, line 19: Is this really necessary? Apparently `devel::m4` does not need it. If I disable `devel::bootstrap-fake-autotools`, I don't get any parsing error either. Do I overlook something? :thinking:

README.md
```diff
@@ -16,7 +16,6 @@ Docs](https://bob-build-tool.readthedocs.io/projects/basement/en/latest/index.ht
   * gcc >= 5.x
   * bash
   * POSIX awk (GNU awk version >= 3.1.5)
-  * GNU m4
```
jkloetzke, line 19: There is `doc/introduction.rst` too...

classes/basement/rootrecipe.yaml
```diff
@@ -38,6 +38,9 @@ depends:
           - name: devel::m4
             use: [tools]
             forward: True
+          - name: devel::help2man
```
jkloetzke, line 41: Could you add it to the list of available tools in `README.md` and `doc/usage.rst`?

classes/basement/rootrecipe.yaml
```diff
@@ -41,6 +41,9 @@ depends:
           - name: devel::help2man
             use: [tools]
             forward: True
+          - name: devel::automake
```
jkloetzke, line 44: Hmm, `automake` is part of the `autotools` tool already. Why do we need it again separately? I would only do this if there is a good reason...

BTW, the "remove automake from host prequisites" commit which adds this as tool dependency, does not really does what the commit message says.

classes/basement/rootrecipe.yaml
```diff
@@ -44,6 +44,9 @@ depends:
           - name: devel::automake
             use: [tools]
             forward: True
+          - name: devel::texinfo
```
jkloetzke, line 47: Could you add it to the list of available tools in `README.md` and `doc/usage.rst`?

classes/basement/rootrecipe.yaml
rhubert: Hmm...looks like this is no longer necessary... These commits are ~4years old I did not clean them up carefully enough...sorry for that...

jkloetzke:
Thanks.

PR #248, opened 2025-02-05
Ocaml

rhubert:
Depends on #247

### DIFF
--- a/classes/ocaml.yaml
+++ b/classes/ocaml.yaml
@@ -1,0 +1,26 @@
+buildTools: [ocaml, ocamlfind]
+buildSetup: |
+    export OCAMLLIB=${BOB_TOOL_PATHS['ocaml']}/../lib
+    export CAML_LD_LIBRARY_PATH=${BOB_TOOL_PATHS['ocaml']}/../lib/stublibs
+    export OCAMLFIND_CONF=${BOB_TOOL_PATHS['ocamlfind']}/../local/etc/findlib.conf
+
+    OCAMLPATH=${OCAMLPATH:-""}
+    OCAMLPATH+="${OCAMLPATH:+:}${BOB_TOOL_PATHS['ocamlfind']}/../lib"
+    for i in "${@:2}" ; do
+        for j in $i/usr/lib/* ; do
+            if [[ -d "$j" ]] ; then
+                if [ -e "$j/META" ] ; then
+                    OCAMLPATH+="${OCAMLPATH:+:}$i/usr/lib"
+                fi
+            fi
+        done
+    done
+
+    [ -z "${OCAMLPATH:+true}" ] || export OCAMLPATH
+
+packageSetup: |
+    ocamlFixupInstall() {
+        find . -type f -exec \
+            sed -i "s@#!.*/ocamlrun@#!/usr/bin/env ocamlrun@g" \
+            {} \;
+    }
--- a/recipes/devel/ocaml.yaml
+++ b/recipes/devel/ocaml.yaml
@@ -1,0 +1,39 @@
+inherit: [autotools]
+
+metaEnvironment:
+    PKG_VERSION: "5.3.0"
+
+checkoutSCM:
+    scm: url
+    url: https://github.com/ocaml/ocaml/archive/refs/tags/${PKG_VERSION}.tar.gz
+    digestSHA256: eb9eab2f21758d3cfb1e78c7f83f0b4dd6302824316aba4abee047a5a4f85029
+    stripComponents: 1
+
+buildTools: [m4]
+buildVars: [STRIP]
+buildScript: |
+    # Note: the configure script is broken - it generates the makefiles in the
+    # directory of configure not in cwd. But the autotools-class expects the makefiles
+    # to be in build. rsync the sources into 'build' dir as workaround
+    mkdir build
+    rsync -a $1/ build/
+
+    # The makefile expects 'strip' in the path...
+    mkdir -p .bob
+    pushd .bob
+    ln -s $(which ${STRIP}) strip
+    popd
+    export PATH=${PATH}:$(pwd)/.bob
+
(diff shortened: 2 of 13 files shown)


classes/ocaml.yaml
```diff
@@ -0,0 +1,25 @@
...
+buildSetup: |
+    export OCAMLLIB=${BOB_TOOL_PATHS['ocaml']}/../lib
+    export OCAMLFIND_CONF=${BOB_TOOL_PATHS['ocamlfind']}/../local/etc/findlib.conf
+
+    OCAMLPATH=${OCAMLPATH:-""}
+    for i in "${@:2}" ; do
+        for j in $i/usr/lib/* ; do
+            if [[ -d "$j" ]] ; then
+                if [ -e "$j/META" ] ; then
+                    OCAMLPATH+=""
```
jkloetzke, line 11: Why add an empty string?

recipes/devel/ocaml.yaml
```diff
@@ -0,0 +1,44 @@
...
+    mkdir -p .bob
+    pushd .bob
+    ln -s $(which ${STRIP}) strip
+    popd
+    export PATH=${PATH}:$(pwd)/.bob
+
+    ./configure \
+            ${AUTOCONF_BUILD:+--build=${AUTOCONF_BUILD}} \
+            ${AUTOCONF_HOST:+--host=${AUTOCONF_HOST}} \
+            ${AUTOCONF_TARGET:+--target=${AUTOCONF_TARGET}} \
```
jkloetzke, line 25: `AUTOCONF_TARGET` is not part of `buildVars`. I guess you have never built a cross-compiler?

recipes/devel/ocaml.yaml
```diff
@@ -0,0 +1,44 @@
...
+buildVars: [STRIP]
+buildScript: |
+    rsync -a --delete $1/ .
+    mkdir -p .bob
+    pushd .bob
+    ln -s $(which ${STRIP}) strip
+    popd
+    export PATH=${PATH}:$(pwd)/.bob
+
+    ./configure \
```
jkloetzke, line 22: Is there a particular reason why you do not use `autotoolsBuild`? This applies to basically all packages.

recipes/devel/opam.yaml
```diff
@@ -0,0 +1,33 @@
+inherit: [autotools]
+
+metaEnvironment:
+    PKG_VERSION: "2.3.0"
+
+checkoutSCM:
+    scm: git
+    url: git@gitlab.lab.dd.secunet.de:hubert/opam.git
```
jkloetzke, line 8: `NXDOMAIN` :wink:

jkloetzke (commented):
Could you please add a chapter in `doc/usage.rst` how Ocaml is supposed to be used? I managed to get a hello world compiled but then it depends on `ocamlrun`. How does that come onto the target? When I tried to use `ocamlopt` I got the following:

```
bash-5.2$ ocamlopt -o hello hello.ml -verbose
+ x86_64-bob_compat-linux-gnu-as  -o 'hello.o' '/tmp/camlasmb63199.s'
sh: line 1: x86_64-bob_compat-linux-gnu-as: command not found
File "hello.ml", line 1:
Error: I/O error: x86_64-bob_compat-linux-gnu-as  -o 'hello.o' '/tmp/camlasmb63199.s'
```

Apparently it needs the original `host-toolchain` which leaves me a bit puzzled. Is this intended?

rhubert:
I'm using this only for host-stuff ATM. In particular some required stuff for compiling mesa with intel drivers is in ocaml. I took a look in how to build a Ocaml cross compiler some months ago, but gave up as there ~is~ was no real out-of-the box support ATM. 

I'll comment this but I don't want to spend time in building a ocaml cross compiler now - if this is ok for you?

For further investigations:
It seams cross compilation support with ocaml 5.3  (released 2025-01-08) has been [improved](https://discuss.ocaml.org/t/building-an-ocaml-cross-compiler-with-ocaml-5-3/15918).
Also [nix](https://github.com/nix-ocaml/nix-overlays/tree/master/cross) builds a ocaml-cross compiler and uses this along with (dune)[https://dune.readthedocs.io/en/latest/cross-compilation.html]

jkloetzke:
> I'll comment this but I don't want to spend time in building a ocaml cross compiler now - if this is ok for you?

This is perfectly fine for me. We can add the cross-compiler support when its needed.

But having some information in the docs about how this is supposed to be used is something that I indeed would want to have. Otherwise it is unusable by others, which would be a pity.

We might also add some simple hello-world test in `tests/linux`...

recipes/devel/ocaml.yaml
rhubert: Yes - because `autotoolsBuild` is not working. Using `autotoolsBuild` it fails with:

 `make: *** No targets specified and no makefile found.  Stop.

The reason is, that `configure` did not produce the makefiles in the current working directory, it's always in the folder of the configure script. - but `autotoolsBuild` uses a build-folder where it runs `configure` and assumes the makefiles to be generated.

To use the code from the `autotools`-class I think we can either
- move the  `configure` part of the `autotools` - class into a separate class and reuse it from there
- add a argument to `autotoolsBuild` skipping the [pushd build](https://github.com/BobBuildTool/basement/blob/master/classes/autotools.yaml#L60)

Do you see another option? Which one do you prefer?

jkloetzke: Oh my gosh! It's really that borken. :facepalm:  In this case I would rather not touch the `autotools` class. But a small comment with exactly this explanation would be appreciated. Otherwise we'll start wondering again a couple of years down the road why the `autotools` class isn't use...

rhubert: I found a solution to use `autotoolsBuild` for most of the packages. It's not pretty but works...

jkloetzke:
Thanks!

PR #249, opened 2025-02-07
python3-usr: provide sepeate class for checkout usage

rhubert:
65cedc9 makes python usabe in checkoutScripts. This produces different checkoutVariants for recipes where the python part is only needed for one multiPackage but not for another:
```
checkoutSCM:
 ...

multiPackage:
    foo:

    python:
        inherit: [python3-usr]
```
Revert 65cedc9 and add a separate python3-usr-checkout class to avoid this.

### DIFF
--- a/classes/python3-usr_checkout.yaml
+++ b/classes/python3-usr_checkout.yaml
@@ -1,0 +1,18 @@
+checkoutTools: [python3]
+checkoutSetup: |
+    export PYTHONNOUSERSITE=1
+
+    if command -v python3 2>&1 >/dev/null; then
+        PYTHONPATH=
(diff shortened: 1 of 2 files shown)


classes/python3-usr_checkout.yaml
```diff
@@ -0,0 +1,17 @@
+checkoutSetup: |
```
jkloetzke, line 1: Should we add a `checkoutTools: [python3]`? I guess there is no point in using this class without the pyhton3 tool...

jkloetzke (commented):
It's unfortunate that we have to duplicate the code but I don't see a better option either...

jkloetzke:
Thanks

PR #251, opened 2025-03-03
Provide texinfo tool and L4Re for ncurses

admlck:


admlck:
Please let me know when patches should be submitted separately.

jkloetzke:
No, I think that's fine. Thanks.

PR #257, opened 2025-04-01
strip: fix stripping of setuid ELF files

rhubert:
The file command reports ELF files with the setuid bit is set as

setuid ELF 64-bit LSB pie executable [...] with debug_info, not stripped

Fix the condition to cover these as well.


rhubert:
(Found for `net::openssh` -> `usr/libexec/ssh-keysign` )

jkloetzke:
Thanks. Nice catch.

PR #262, opened 2025-05-26
openssl and pkg-config fix

sixtyfourktec:


classes/install.yaml
```diff
@@ -41,6 +41,11 @@ packageSetup: |
                 echo "$K"
                 continue
             fi
+            # Skip values which are a comment
+            if [[ $V =~ [:space:]*#+.* ]] ; then
+                echo "$K"
+                continue
+            fi
```
jkloetzke, line 48: I think we should always strip comments as the first step. Otherwise something like

```sh
Libs = foo # somment
Libs.private = bar
```

will still fail. So what about the following?

```sh
        while IFS=: read -r K V ; do
            V="${V%%#*}"
...
```

sixtyfourktec: done

Issue #268, opened 2025-08-15
sandbox not compiling with host gcc15


rhubert:
If the initial host compiler is `gcc15` it's not possible to compile a new sandbox image since `gcc15` has `c23` enabled by default. This breaks (at least) `pkg-config`, `binutils`, `m4`, `libgmp`. 

I'm working on updating the affected packages - or do we simply want to add `-std=c18` to the default `CFLAGS`?

PR #270, opened 2025-08-22
Perl+c23

rhubert:
The reason for this change is to add support for perl modules as needed for example by graphviz (see https://github.com/BobBuildTool/basement-gnu-linux/issues/5)
It seams to be crucial to have the same environment when compiling perl and the perl module, which is not the case if perl is in the sandbox / the host perl and the module is build using the host-compat-toolchain. By using perl as a regular tool perl modules can be used like python modules.

As a new sandbox is necessary I run into  https://github.com/BobBuildTool/basement/issues/268. The fixes are also added here for now. I can separate the PR's if you want to.

### DIFF
--- a/classes/perl5/build.yaml
+++ b/classes/perl5/build.yaml
@@ -1,0 +1,14 @@
+inherit: [perl5, 'basement::bits::libs']
+
+depends:
+    - name: perl::perl-dev
+      use: [result, deps, environment]
+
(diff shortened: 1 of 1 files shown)


rhubert:
Thanks for the feedback :+1: I'll rework it following your suggestions but it might take some time to recompile everything...

PR #308, opened 2026-03-06
python3::build: add option to build the wheel only

rhubert:
Without any additional option both sdist and wheel are generated.


jkloetzke:
Thanks

PR #316, opened 2026-04-24
strip: fix strip of shared libraries

rhubert:
The file command returns `ELF 64-bit LSB shared object, x86-64, [..]` for shared libraries. Fix the pattern to match this. Otherwise shared libraries might miss some symbols.


jkloetzke:
Thanks

PR #319, opened 2026-08-25
net/make-ca: fix cross-compilation

mahaase:
fix for the problem, that i get the error, that `/usr/bin/openssl` is not found while build.

### DIFF
--- a/recipes/net/make-ca.yaml
+++ b/recipes/net/make-ca.yaml
@@ -12,9 +12,25 @@ checkoutSCM:
         digestSHA1: b7be1905a9ca11c0c42d0270ce54132f354244b3
         dir: nss
 
+depends:
+    - tools:
+          target-toolchain: host-compat-toolchain
+      depends:
+          - name: libs::openssl-tool
+            use: [tools]
+
 buildScript: |
     cp $1/make-ca/make-ca .
     cp $1/nss/certdata.txt .
 
+    # Use 'openssl rehash' instead of 'c_rehash', cherry-pick of commit (v0.8):
+    # https://github.com/djlucas/make-ca/commit/486284c5ad9a9d0018473eb8f4baacb43a397cfe
+    grep -q '^/usr/bin/c_rehash ' make-ca
+    sed -i -e 's|^/usr/bin/c_rehash |"${OPENSSL}" rehash |' make-ca
+
+packageTools: [openssl]
 packageScript: |
-    bash $1/make-ca -C $1/certdata.txt -D "$PWD"
+    # make-ca has no auto detection but defaults to /usr/bin/openssl. Point it
+    # at libs::openssl-tool so that the result does not depend on the host.
+    OPENSSL_BIN="${BOB_TOOL_PATHS[openssl]}/openssl"
+    bash $1/make-ca -C $1/certdata.txt -D "$PWD" -s "$OPENSSL_BIN"


jkloetzke:
I've restarted the Jenkins builder. It somehow got stuck.

Hmm, the requirement to set the `ARCH` is unexpected. In your quote (from $AITOOL I guess):

> ARCH has to be set explicitly because the host-compat-toolchain does not provide it

This is not the case in the master branch. So something else must be going wrong...

mahaase:
> if i build the generic sandbox. 

this just happens in combination with the generic sandbox feature. have you tested this in exactly this case?

jkloetzke:
No, I just looked at the change. Could you spend some tokens to find out why it is required to override `ARCH`? The `host-compat-toolchain` provides `ARCH` as tool variable. In theory, the env override should be ignored. I think we have to understand this before merging this.

recipes/net/make-ca.yaml
```diff
@@ -12,9 +12,27 @@ checkoutSCM:
...
+            use: [tools]
+
 buildScript: |
     cp $1/make-ca/make-ca .
     cp $1/nss/certdata.txt .
 
+    # Use 'openssl rehash' instead of 'c_rehash', cherry-pick of commit (v0.8):
+    # https://github.com/djlucas/make-ca/commit/486284c5ad9a9d0018473eb8f4baacb43a397cfe
+    grep -q '^/usr/bin/c_rehash ' make-ca
+    sed -i -e 's|^/usr/bin/c_rehash |"${OPENSSL}" rehash |' make-ca
```
jkloetzke, line 29: Hmm, where is the `OPENSSL` variable defined?

mahaase: This is a `make-ca`'s own variable, not a variable of the build step. the sed expression is single quoted and not expanded at build time. `make-ca` handles it, where we set the value via `-s "$OPENSSL_BIN"` on lines 37-38.

jkloetzke: Ah, makes sense. :+1:

jkloetzke:
Did #323 solve the problem with the unexpected `ARCH` variable?

mahaase:
> No, I just looked at the change. Could you spend some tokens to find out why it is required to override `ARCH`? The `host-compat-toolchain` provides `ARCH` as tool variable. In theory, the env override should be ignored. I think we have to understand this before merging this.

ofc, a little too detailed for me 🤯:

> ### Why ARCH had to be overridden
>
> Your assumption holds for `devel::host-compat-toolchain`, which indeed provides `ARCH: "${ARCH}"` in
> its tool environment. But while the sandbox itself is built, that recipe is not in scope.
> `devel::sandbox-default` gets its tools from `devel::bootstrap-sandbox-toolchain`, which inherits
> `basement::bits::sandbox-toolchain`, and there `host-compat-toolchain` is just an alias of the
> sandbox's own native toolchain:
>
> ```
> classes/basement/bits/sandbox-toolchain.yaml:52   target-toolchain: &target-toolchain
> classes/basement/bits/sandbox-toolchain.yaml:83   host-compat-toolchain: *target-toolchain
> ```
>
> That tool environment defines `CC`/`CXX`/… and `TOOLCHAIN_FLAVOUR` only — no `ARCH`. So there is no tool
> variable that could override anything and `libs::openssl-tool` takes `ARCH` straight from the recipe
> environment; `ARCH` is in openssl's `buildVars`, the configure target is derived from it.
>
> And in the generic-sandbox case that environment value describes the wrong machine: the class now
> provides `ARCH: "${SANDBOX_ARCH:-$ARCH}"` via `provideVars`, and `provideVars` are substituted with
> the package step's environment. devel::bootstrap-sandbox-toolchain pins only
> `SANDBOX_AUTOCONF_BUILD` in its `privateEnvironment`, while `SANDBOX_ARCH` is inherited unchanged
> from `devel::sandbox-default`. So the bootstrap toolchain — x86_64, running in the x86_64 bootstrap
> sandbox — advertises `ARCH=arm64` and forwards it to every subsequent dependency of
> `devel::sandbox-default`, `net::make-ca` included. openssl is then configured as `linux-aarch64` but
> compiled with the x86_64 compiler, hence `#error "unsupported ARM architecture"`. On master
> `SANDBOX_ARCH` is unset, so the forwarded ARCH happens to be the host arch and nothing breaks.
>
> ### Better fix
>
> `ARCH: "$(host-arch)"` here only papers over that, and it is wrong in general: if the whole build
> runs inside a foreign sandbox, the tool has to match the sandbox arch, not the build machine. I'd
> rather fix it in the generic-sandbox change:
>
> 1. Pin `SANDBOX_ARCH` per sandbox toolchain, symmetric to the existing `SANDBOX_AUTOCONF_BUILD`:
>    `SANDBOX_ARCH: "$(host-arch)"` in `devel::bootstrap-sandbox-toolchain` and
>    `SANDBOX_ARCH: "${SANDBOX_ARCH:-$(host-arch)}"` in `devel::sandbox-toolchain`, so each toolchain
>    describes the sandbox it belongs to. Target packages are unaffected, they get `ARCH` from the
>    cross toolchain's tool environment (`recipes/devel/cross-toolchain.yaml:95`).
> 2. Additionally add `ARCH: "${SANDBOX_ARCH}"` to the tool environment of the `&target-toolchain`
>    anchor in `basement::bits::sandbox-toolchain`, so `host-compat-toolchain` behaves like
>    `devel::host-compat-toolchain` and really overrides the environment — that also makes your
>    original expectation true. No new tool-variable clashes: the anchor already defines `CC`/`AR`/…, so
>    any step using two of those aliases at once would fail today already.
>
> Then the `environment: ARCH:` block can be dropped from this PR; the rest (`host-compat-toolchain` for
> openssl-tool, `packageTools`, `-s`, the c_rehash sed) is needed independently of the generic
> sandbox, otherwise the package step silently uses the host's `/usr/bin/openssl`.
>
> Side note: the sandbox alias also does not reset CPPFLAGS/CFLAGS/CXXFLAGS/LDFLAGS the way
> devel::host-compat-toolchain does, so target flags (-mcpu/-march, if CROSS_TOOLCHAIN_CPU/
> CROSS_TOOLCHAIN_ARCH are set) leak into the host build of openssl-tool.

If I understand correctly, this change here is wrong.
So I'll drop the `ARCH` override here and put 1. & 2. into the generic-sandbox PR and i'll integrate #323.

Everything else inside this PR is needed in my opinion.

jkloetzke:
Thanks. Indeed, with the above explanation, the `ARCH` override is now unnecessary with the fix in #323.

PR #321, opened 2026-09-04
devel/compat/gcc: pass gmp/mpfr/mpc/isl paths to configure

### LINKED ISSUE #320: devel::host-compat-toolchain fails for aarch64 targets
devel::host-compat-toolchain fails for aarch64 targets: gmp.h not found (GCC t-aarch64 ALL_SPPFLAGS typo)

## Summary

`devel::host-compat-toolchain` cannot be built whenever its isolation toolchain targets an aarch64 machine. The build dies in `devel::compat::gcc-cross-bare`:

```
In file included from gcc/config/aarch64/aarch64-speculation.cc:22:
gcc/system.h:687:10: fatal error: gmp.h: No such file or directory
  687 | #include <gmp.h>
make[2]: *** [gcc/config/aarch64/t-aarch64:81: aarch64-speculation.o] Error 1
```

Failed package:
`devel::host-compat-toolchain/devel::compat::cross-toolchain/devel::compat::gcc-cross/libs::compat::glibc/devel::compat::gcc-cross-bare`

## Root cause

`gcc/config/aarch64/t-aarch64` has a typo in the rule that builds `aarch64-speculation.o`:

```make
        $(COMPILER) -c $(ALL_COMPILERFLAGS) $(ALL_SPPFLAGS) $(INCLUDES) \
                                            ^^^^^^^^^^^^^^
          $(srcdir)/config/aarch64/aarch64-speculation.cc
```

Every other rule in that very same file uses `$(ALL_CPPFLAGS)`. `ALL_SPPFLAGS` is undefined, so it expands to nothing and, because
`ALL_CPPFLAGS = $(INCLUDES) $(CPPFLAGS)`, the whole `$(CPPFLAGS)` part is silently dropped from that one compile line.

`recipes/devel/compat/gcc.yaml` locates gmp/mpfr/mpc/isl by appending `-I` flags to `CPPFLAGS` (see the loop over `$2`..`$5` in its `buildScript`), not by passing `--with-gmp-include` & friends to configure. Those include paths therefore never reach `aarch64-speculation.cc`.

Verified present in GCC **9.5.0** (the version `devel::compat::gcc` uses) and in GCC **11.4.0**. It is also still present on current GCC master, where it has meanwhile been copy-pasted into a second rule (`aarch64-elf-metadata.o`), so an upstream GCC fix does not look imminent.

## Who is affected

* **Native aarch64 build hosts.** `recipes/devel/host-compat-toolchain.yaml` configures the isolation toolchain with `AUTOCONF_TARGET: "$(gen-autoconf,bob_isolation)"`, and `$(gen-autoconf,...)` derives the machine from the build machine. On an arm64 host that is `aarch64-bob_isolation-linux-gnu`, so `t-aarch64` is used and the build breaks. As far as I can tell `devel::host-compat-toolchain` — and therefore `basement::rootrecipe` — simply cannot be built on aarch64 today.
* Any setup where the sandbox architecture is aarch64 (this is how we ran into it).

## Why nobody hit this so far

* With an x86_64 target, GCC uses `config/i386/t-i386` and never reads `t-aarch64` at all.
* `recipes/devel/gcc.yaml` (the non-compat GCC) has the identical typo in its 11.4.0 source but is unaffected, because it passes `--with-gmp-include=...` to configure. That lands in `GMPINC`, which is part of `$(INCLUDES)` — and `$(INCLUDES)` *is* present on the broken compile line.

## Two possible fixes

1. **Patch the GCC source.** Add a patch to the `devel::compat::gcc` series that turns `$(ALL_SPPFLAGS)` into `$(ALL_CPPFLAGS)`. One-line, obviously
   correct, applies with zero fuzz.
2. **Change the recipe.** Make `recipes/devel/compat/gcc.yaml` pass `--with-gmp-include`/`--with-gmp-lib` (and mpfr/mpc/isl) like `recipes/devel/gcc.yaml` already does, instead of relying on `CPPFLAGS`. More invasive, but it removes the dependency on a GCC makefile detail and would also fix the linker side.

* Which of the two would you prefer for basement?
* If it is the patch: should the same patch also be added to the `recipes/devel/gcc/` series for consistency, even though that GCC is not affected in practice? That would invalidate the cached artifacts of every GCC build.
* Note that either fix changes the variant-id of `devel::compat::gcc` and thus forces a one-time rebuild of the compat toolchain on x86_64 as well (bit-identical result, since `t-aarch64` is not used there).

mahaase:
Building `devel::compat::gcc-cross-bare` for aarch64 targets failed with:
```
    In file included from gcc/config/aarch64/aarch64-speculation.cc:22:
    gcc/system.h:687:10: fatal error: gmp.h: No such file or directory
```

The rules in `gcc/config/aarch64/t-aarch64` reference `$(ALL_SPPFLAGS)` instead of `$(ALL_CPPFLAGS)`. The undefined variable expands to nothing so the include paths that were passed via `CPPFLAGS` are dropped for these objects. The typo is still present in current GCC master.

Those rules do keep `$(INCLUDES)`, which contains `$(GMPINC)` and `$(ISLINC)`. Therefore locate the libraries the same way as the regular `gcc` recipe does, that is by telling configure explicitly where the headers and libraries are. This also removes a needless deviation between the two recipes.

Fixes #320

### DIFF
--- a/recipes/devel/compat/gcc.yaml
+++ b/recipes/devel/compat/gcc.yaml
@@ -74,7 +74,15 @@ buildScript: |
                 --disable-documentation \
                 --disable-debug \
                 --with-xmlto=no \
-                --with-fop=no
+                --with-fop=no \
+                --with-gmp-include=${BOB_DEP_PATHS[libs::gmp-dev]}/usr/include \
+                --with-gmp-lib=${BOB_DEP_PATHS[libs::gmp-dev]}/usr/lib \
+                --with-mpfr-include=${BOB_DEP_PATHS[libs::mpfr-dev]}/usr/include \
+                --with-mpfr-lib=${BOB_DEP_PATHS[libs::mpfr-dev]}/usr/lib \
+                --with-mpc-include=${BOB_DEP_PATHS[libs::mpc-dev]}/usr/include \
+                --with-mpc-lib=${BOB_DEP_PATHS[libs::mpc-dev]}/usr/lib \
+                --with-isl-include=${BOB_DEP_PATHS[libs::compat::isl-dev]}/usr/include \
+                --with-isl-lib=${BOB_DEP_PATHS[libs::compat::isl-dev]}/usr/lib
             touch .configure.stamp
         fi
         makeParallel


jkloetzke:
Thanks!

PR #323, opened 2026-09-07
Fix host native cflags

jkloetzke:
Not all host-compat-toolchain instances and none of the host-native-toolchain's did reset the C/CPP/CXX/LD-FLAGS. This can lead to subtly different variants because some state of the current cross compiling toolchain leaks when pivoting to the compat toolchain.

### DIFF
--- a/classes/basement/bits/sandbox-toolchain.yaml
+++ b/classes/basement/bits/sandbox-toolchain.yaml
@@ -73,6 +73,7 @@ provideTools:
             GNU_LD: "ld"
 
             # meta information
+            ARCH: "${ARCH}"
             TOOLCHAIN_FLAVOUR: gcc
         fingerprintIf: True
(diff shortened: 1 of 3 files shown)


mahaase:
ouh! I guess these changes are related the the generic sandbox build feature too?! I will merge it into local test setup.

jkloetzke:
> ouh! I guess these changes are related the the generic sandbox build feature too?! I will merge it into local test setup.

Actually, I found this problem on a different project (https://github.com/L4Re/l4re-tutorials) where the following dependency did not download the toolchain as expected but started to build it:

```yaml
depends:
  - tools:
      target-toolchain: host-compat-toolchain
    depends:
      - name: devel::cross-toolchain-aarch64-none-elf
        use: [tools, environment]
```

But it could very well explain the make-ca problem too.

PR #328, opened 2026-09-10
devel::binutils: Enable specification of targets

rhubert:
The new BINUTILS_TARGETS_BFD variable can be used to build libbfd with support for additional targets.
The BINUTILS_ENABLE_64BIT_BFD boolean variable can be used to explicitly include support for 64 bit targets. While it is enabled by default when building for 64 bit targets, it is disabled when setting BINUTILS_TARGETS_BFD to all.


recipes/devel/binutils.yaml
```diff
@@ -50,6 +60,8 @@ multiPackage:
                 --disable-werror \
                 --enable-plugins \
                 --enable-lto \
+                ${BINUTILS_TARGETS_BFD:+--enable-targets=${BINUTILS_TARGETS_BFD}} \
+                $( [ $BINUTILS_ENABLE_64BIT_BFD ] && echo "--enable-64-bit-bfd" || echo "" ) \
```
jkloetzke: This will *always* set `--enable-64-bit-bfd`. The `test` builtin will treat any non-empty string as "true". 

```suggestion
                $( (( $BINUTILS_ENABLE_64BIT_BFD )) && echo "--enable-64-bit-bfd" ) \
```

rhubert: :see_no_evil: Thanks.

jkloetzke:
LGTM. Will be merged with the other queued PRs.

jkloetzke:
Thanks!

PR #329, opened 2026-09-15
meson: fix meson_join_list with empty flags array

### LINKED ISSUE #325: meson_join_list fails with $1: unbound variable
`meson_join_list fails` with "$1: unbound variable" when a flags array is empty in `classes/meson.yaml`.

## Summary

Every meson-based build whose `LDFLAGS` (or `CFLAGS`/`CXXFLAGS`) are empty prints spurious
`unbound variable` errors to the build log. The messages point at an unrelated line number and
recipe, which makes them look like a defect in the recipe being built.

## Observed behaviour

```
   BUILD     work/libs/jsoncpp-dev/aarch64-bob_compat-linux-gnu/O2pipefPIC/build/2/workspace
/.script: line 195: $1: unbound variable
/.script: line 195: $1: unbound variable
```

The build itself completes successfully.

## Steps to reproduce

1. Build any recipe inheriting the `meson` class for a configuration where `LDFLAGS` is unset or
   empty, e.g. `bob build libs::jsoncpp-dev` for `aarch64-bob_compat-linux-gnu` / `O2pipefPIC`.
2. Observe two `$1: unbound variable` messages on stderr during the build step.

## Root cause

`classes/meson.yaml:12` defines a single-line helper that unconditionally dereferences `$1`:

```bash
function meson_join_list { echo -n "'$1" ; shift ; printf "%s" "${@/#/\', \'}"; echo "'" ; }
```

It is called at `classes/meson.yaml:90-93` while generating `cross_file.txt`:

```bash
c_link_args   = [$(meson_join_list "${MESON_LDFLAGS[@]}")]
cpp_link_args = [$(meson_join_list "${MESON_LDFLAGS[@]}")]
```

`MESON_LDFLAGS` is built via `declare -a MESON_LDFLAGS=( ${LDFLAGS:-} )` (`meson.yaml:19`). When
`LDFLAGS` is empty the array has zero elements, so `"${MESON_LDFLAGS[@]}"` expands to zero words and
the function is invoked with no arguments. Bob runs step scripts under `set -u`, so referencing `$1`
aborts. Bash attributes the error to the line containing the expansion — the one-line function
definition — hence the misleading `line 195`, which is neither the caller nor anything in the
recipe. Two calls (`c_link_args`, `cpp_link_args`) produce exactly two messages.

## Why the build still succeeds

The calls sit inside a command substitution, so `set -u` terminates only the subshell. The
substitution yields an empty string and the generated `cross_file.txt` ends up correct by accident:

```
c_args = ['-O2', '-pipe', '-fPIC']
c_link_args = []
cpp_link_args = []
```

## Impact

- Noise in build logs; misleading line number and false attribution to the recipe under build.
- Latent risk: any future call to `meson_join_list` outside a command substitution would abort the
  whole build step with the same unhelpful message.

## Suggested fix

Handle the zero-argument case in `classes/meson.yaml:12`:

```bash
function meson_join_list { (( $# )) || return 0; echo -n "'$1" ; shift ; printf "%s" "${@/#/\', \'}"; echo "'" ; }
```

`meson_join_space` (`meson.yaml:11`) is unaffected — `"$*"` is safe with zero arguments.

mahaase:
meson_join_list unconditionally dereferenced $1. Bob runs step scripts with "set -u", so calling it with an empty flags array (e.g. an unset or empty LDFLAGS) aborted with "$1: unbound variable". The error was attributed to the line of the one-line function definition, making it look like a defect of the recipe that was built.

The message was harmless so far because all call sites are inside a command substitution: only the subshell died and the substitution expanded to the empty string, which happens to be the correct result. Handle the zero argument case explicitly instead.

Fixes #325


jkloetzke:
LGTM. Will be merged with the other basement PRs.

jkloetzke:
Thanks!

PR #330, opened 2026-09-15
Bump compat

rhubert:
Ubuntu 22.04 is the oldest supported release (standard support). Bump host-compat* to the versions used in this release.


mahaase:
I'm curious about this! 
We are still using Ubuntu 22 LTS, becuase of autoware/ros2 (humble), also with NVIDIA AGX Orin Jetpack 6.2.2.

my targets `gcc --version`'s:
`gcc (Ubuntu 11.4.0-1ubuntu1~22.04.3) 11.4.0`

https://packages.ubuntu.com:
```
[jammy (22.04LTS)](https://packages.ubuntu.com/jammy/gcc) (devel): GNU C compiler
4:11.2.0-1ubuntu1: amd64 arm64 armhf i386 ppc64el riscv64 s390x
```

everything else is fine.

rhubert:
Sorry - I did not get what you want to say - and I'm not aware of any of these `autoware/ros2 (humble) [..] NVIDIA AGX Orin Jetpack 6.2.2.`  buzzwords.... :shrug: Do you need some time for testing this?

gcc12 is available in ubuntu 22.04 : https://packages.ubuntu.com/jammy/gcc-12 ? I guess you can just install this package and everything should work fine? I'd like to use gcc12 because we have packages failing with gcc < 12

mahaase:
> I did not get what you want to say 

just tried to say, that gcc 12 is not the default gcc on ubuntu 22 lts 🤯

but i did some research and i guess you are right. i was afraid, because cuda compilers are strongly restricted to specific gcc versions. but gcc12 is supported. 

> gcc12 is available in ubuntu 22.04 

right. i failed.

go for it! 😊

rhubert:
> > I did not get what you want to say
> 
> just tried to say, that gcc 12 is not the default gcc on ubuntu 22 lts 🤯

I noticed this after your comment :) Since I was to lazy I gave my agent the task to update the compat-packages to match ubuntu 22.04... Not sure why gcc12 was the result ... 

> because cuda compilers are strongly restricted to specific gcc versions

Out of curiosity: this is host-compat. Why do you need this toolchain for your cuda (cross?) compiler? :thinking:

mahaase:
maybe i had just in mind, that the compat version shall not be higher as the normal gcc.
we already forked the basement because of the ubuntu 22 (jetpack 6.2.2) target systems, for which we build our stuff.
so currently our normal gcc is gcc 11 too.
but in the end i want to keep the necessary diff as small as possible.

recipes/devel/host-compat-toolchain.yaml
```diff
@@ -109,7 +109,7 @@ fingerprintScript: &check-host |
         echo "These recipes assume that your host is using libstdc++!" >&2
         exit 1
     fi
-    if [[ $STAMP -lt 20220313 ]] ; then
+    if [[ $STAMP -lt 20230526 ]] ; then
```
jkloetzke: These Ubuntu guys are driving me crazy. The default compiler on Ubuntu 22.04 is gcc-11. This installs libstdc++-11-dev, which has:

```c
#define __GLIBCXX__ 20230528
```

Interestingly, libstdc++6, the actual shared library, comes from gcc-12. And in libstdc++-12-dev, the definition is:

```c
#define __GLIBCXX__ 20230510
```

Do you remember where the definition came from?

rhubert: :/ unfortunately not... :thinking: I did not find any release with this date...maybe just fat-fingers... I haven't tested this on a ubuntu 22.04 system...

But as you pointed out the timestamp of the version in gcc12 is lower than the timestamp of the one gcc11. So this test seams to be broken anyway?  

I spend some token the get a potentially better solution. It compares / checks the `GLIBCXX_*` and `GLIBCXX_ABI_*`  versions instead of the date.  Test on Debian 13.7 / gcc 14.2.0 was successful..

recipes/devel/host-compat-toolchain.yaml
```diff
@@ -98,19 +98,58 @@ fingerprintScript: &check-host |
...
+    # stamp, so it cannot reliably express this ABI requirement.
+    HOST_LIBSTDCXX="$(g++ -print-file-name=libstdc++.so.6)"
+    if [[ ! -f $HOST_LIBSTDCXX ]] ; then
+        echo "Could not locate the host libstdc++.so.6!" >&2
         exit 1
     fi
-    if [[ $STAMP -lt 20220313 ]] ; then
-        echo "Your host libstdc++ version is too old!" >&2
+
+    COMPAT_LIBSTDCXX="$(find /toolchain/sysroots -name 'libstdc++.so.6*' -type f -print -quit)"
```
jkloetzke: I'm afraid that doesn't work. This path only exists inside the sandbox where the test as a whole is unnecessary. The real reason for the finger print script is for builds without sandbox. An that is going to bail out below now.

tests/unit/test-host-compat-toolchain-fingerprint.sh
jkloetzke, line 1: I don't think such a test is really necessary. I also find it very brittle to extract the finger print script with awk from a yaml file. I'd just remove it.

recipes/devel/host-compat-toolchain.yaml
jkloetzke: But I think the idea is still worth to be explored. I'd keep the way how libstdc++.so is found and just do the following test:

```sh
if ! strings "$HOST_LIBSTDCXX" | grep -q '^GLIBCXX_3.4.30$' ; then
    echo "Your host libstdc++ version is too old!" >&2
    exit 1
fi
```

(based on libstdc++.so.6 that I just downloaded from packages.ubuntu.com for Ubuntu jammy)

rhubert: I changed `grep -q` to `grep .. > /dev/null`. Otherwise grep exits on the first match while string still produces output resulting an a `SIGPIPE`....

jkloetzke: :exploding_head: Indeed. This is caused by `set -o pipefail` in Bob. Good catch!

jkloetzke:
Thanks!

PR #331, opened 2026-09-15
opam: bump to 2.5.2 opam-full version

mahaase:
The recorded digest does not match the upstream release tarball, so the checkout fails. Update it to the actual sha256 of
ocaml/opam 2.3.0.tar.gz.


rhubert:
I think we should switch to the `opam-full` package. This is not generated from the tag and should be stable :middle_finger: . The difference between these packages are additionally dependencies bundled in `src_ext`.

mahaase:
i will try to switch this.

mahaase:
i tested the full-opam version incl. bump to 2.5.2.
worked out of the box in my tests. nothing else to adapt.

recipes/devel/opam.yaml
rhubert: As you are on it: please add the missing `-p` :)

jkloetzke:
LGTM. Will be merged with the other pending PRs.

jkloetzke:
Thanks!

PR #333, opened 2026-09-16
libs/libpcre2: unify with basement-gnu-linux libs/pcre2

MarcKe:
I have noticed the two version of pcre2. There does not seem to be a necessity to have both.


jkloetzke:
LGTM. Will be merged with the other pending PRs.

jkloetzke:
Thanks!

